PR #41, opened 2025-12-11
Django atomic

mitch-mitchel:
This PR is a major refactor to the django-cotton/AlpineJS/Tailwind stack that matches the new stack for CourtListener. It mirrors some of those existing best practices, and adds an Atomic Design Structure to the Django-Cotton templates and a custom living style guide. This is the first phase of UX/Design to get to the goal of a clickable Demo by the ITC conference in January.

### DIFF
--- a/templates/pages/components.html
+++ b/templates/pages/components.html
@@ -499,10 +499,10 @@
                                         {# Demo #}
                                         <h4 class="mb-3">Demo</h4>
                                         <div class="space-y-4 mb-6">
-                                          <c-alert variant="info">This is an info alert.</c-alert>
-                                          <c-alert variant="success">This is a success alert.</c-alert>
-                                          <c-alert variant="warning">This is a warning alert.</c-alert>
-                                          <c-alert variant="danger">This is a danger/error alert.</c-alert>
+                                          <c-atoms.alert variant="info">This is an info alert.</c-atoms.alert>
+                                          <c-atoms.alert variant="success">This is a success alert.</c-atoms.alert>
+                                          <c-atoms.alert variant="warning">This is a warning alert.</c-atoms.alert>
+                                          <c-atoms.alert variant="danger">This is a danger/error alert.</c-atoms.alert>
                                         </div>
                                         {# Props #}
                                         <h4 class="mt-8">Props</h4>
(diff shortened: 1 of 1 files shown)


static/css/main.css
elisa-a-v: This file feels a little bigger than what I've seen is standard TW. Reuse is great, but too many `@apply` wrappers become a second styling system: it adds indirection and coupling, which makes it harder to review and change safely than keeping most styling in components/templates with Tailwind tokens.

This is of course a matter of preference but TW strongly suggests this, and my experience has been much nicer than expected so far. I would suggest we consider avoiding utility wrappers unless strictly necessary, and adding them only once they're actually needed (so we keep YAGNI in mind 😉)

templates/cotton/atoms/button.html
```diff
@@ -0,0 +1,8 @@
+<c-vars type="button" variant="primary" size="" />
+<button type="{{ type }}"
+        {% if id %}id="{{ id }}"{% endif %}
+        {% if disabled %}disabled{% endif %}
+        {% if aria_label %}aria-label="{{ aria_label }}"{% endif %}
+        class="{% if variant == 'primary' %}btn-primary{% elif variant == 'outline' %}btn-outline{% elif variant == 'dark' %}btn-dark{% elif variant == 'ghost' %}btn-ghost{% elif variant == 'danger' %}btn-danger{% else %}btn-primary{% endif %}{% if size == 'xl' %} btn-xl{% elif size == 'sm' %} btn-sm{% endif %}{% if full_width %} w-full{% endif %} {{ class }}"
```
elisa-a-v: Neither `class` nor `full_width` are declared in the `c-vars` tag, but I would suggest we reconsider using `full_width`, `variant` and `size` since they are redundant with `class`, what do you think?

templates/cotton/atoms/icon.html
elisa-a-v: this component expects certain attributes but doesn't have a `<c-vars>` tag.

elisa-a-v (commented):
I went over the code and this is a great PR, so thorougly documented, thank you for that! Amazing work getting started with an unfamiliar stack, big props to you 🎉 

Still need to do a proper review but I'm posting some comments I have so far for us to discuss in our meeting later today :)

static/css/main.css
mitch-mitchel: great catch on this one, I misunderstood the @apply. I just want to the styles to be local to the component, which still allows for atomic composition

PR #51, opened 2025-12-29
Layout changes

mitch-mitchel:
This is a cleanup commit that adds some functional tests, sets `pre-commit` as the proper format/linter, and fixes the layout to focus on the AI/chat interface.


mitch-mitchel:
this PR is being broken up into more reviewable chunks

PR #57, opened 2026-01-01
Alpine local deps

### LINKED ISSUE #46: Use local packages for AlpineJS & Tailwind CSS for building
Currently using CDNs for the above, but should build the app from local versions as the CDN not being available can cause issues. This assures the dev environment works all the time, and helps with version control.

mitch-mitchel:
This PR fixes #46
Provides local AlpineJS instead of CDN, both minified and source map versions
Temporarily removes CSP versions for the ITC demo due to chat issues (See https://github.com/freelawproject/litigant-portal/issues/58)


mlissner:
I'm glad to see this. In general, we don't ever want remote sources pulled in. They're not faster (we'll have our own CDN), they introduce another point of failure (their CDN), and they aren't as private. Guessing you know as much, but wanted to note that I have a strong aversion to CDN-hosted JS/CSS stuff.

PR #112, opened 2026-02-18
Add about, privacy, accessibility pages and fix scroll

mitch-mitchel:
## Summary
  - Add `/about/`, `/privacy/`, `/accessibility/` pages with views, routes, and tests
  - Wire footer links to real pages instead of `#` anchors; link FLP name to free.law
  - Fix base layout: `h-dvh` → `min-h-dvh` so content pages scroll naturally
  - Fix header sticky default: empty string is falsy in Django templates, `"false"` is not
  - Replace PostgreSQL with SQLite everywhere (drops ~$31/mo Fly Postgres)
  - Add configurable `CHAT_MODEL` setting via env var
  - Add stale issue/PR workflow

  ## Note
  Chat page scroll may need a follow-up fix — the chat message area does not scroll when
  content overflows. The `h-dvh flex overflow-hidden` body class pins the layout correctly,
  but `chat-messages-area` may need explicit overflow handling.

  ## Test plan
  - [x] `/about/`, `/privacy/`, `/accessibility/` return 200 with correct content
  - [x] Footer links navigate to real pages (not `#` anchors)
  - [x] Content pages scroll naturally (header not sticky)
  - [x] `/chat/` header stays pinned, verify chat messages area scrolls
  - [x] `make test` passes

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -25,7 +25,7 @@
 
       </head>
       <body class="{% block body_class %}
-         h-dvh flex flex-col bg-greyscale-25{% endblock body_class %}
+         min-h-dvh flex flex-col bg-greyscale-25{% endblock body_class %}
          ">
         {% block header %}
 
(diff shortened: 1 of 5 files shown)


nadahlberg (approved):
Looks good to me!

Issue #101, opened 2026-02-11
Add I18n django translation text


mitch-mitchel:
See: https://docs.djangoproject.com/en/6.0/topics/i18n/translation/
Top-level docs: https://docs.djangoproject.com/en/6.0/topics/i18n/

This should be done as early as possible, as it becomes more difficult as time goes on.

## Implementation Plan

Stacked PRs, one per issue. Each depends on the one above it.

- [ ] #143 — i18n infrastructure (settings, middleware, URLs, locale dir)
- [ ] #144 — Python string marking (models, forms, views)
- [ ] #145 — Template strings: static pages & auth
- [ ] #146 — Template strings: chat & Cotton components
- [ ] #147 — JavaScript string translation (JavaScriptCatalog + gettext)
- [ ] #148 — Spanish extraction + translation workflow docs

### Key decisions
- **Cotton props**: Translate at call site using `{% trans "x" as var %}` pattern (djlint-safe)
- **JS**: Django's `JavaScriptCatalog` (CSP-safe, no eval)
- **URLs**: `i18n_patterns` for portal only; API/admin/accounts stay prefix-free
- **Court partners**: `LOCALE_PATHS` ordering for per-deployment overrides + Weblate for non-dev contributors

PR #150, opened 2026-02-27
i18n: add translation infrastructure

mitch-mitchel:
## Summary
- Add `LANGUAGES` (en, es), `LOCALE_PATHS`, `LocaleMiddleware` to settings
- Wrap `portal.urls` in `i18n_patterns` with `prefix_default_language=False`
- Add `JavaScriptCatalog` and `set_language` endpoints
- Update `base.html`: `{% load i18n %}`, dynamic `lang` attr, jsi18n script
- Create `locale/` directory

Part of #101 — PR 1 of 6

## Test plan
- [ ] `make lint` passes
- [ ] `make test` passes
- [ ] All existing URLs work without `/en/` prefix

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -80,6 +80,7 @@ MIDDLEWARE = [
     "django.middleware.security.SecurityMiddleware",
     "csp.middleware.CSPMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
+    "django.middleware.locale.LocaleMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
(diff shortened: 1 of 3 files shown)


mitch-mitchel:
Clean, standard Django i18n setup. `LocaleMiddleware` is in the correct position (after `SessionMiddleware`, before `CommonMiddleware`). `i18n_patterns` with `prefix_default_language=False` is the right call — English URLs stay at their existing paths, `/es/` prefix only added for Spanish.

One suggestion: a short comment in `urls.py` explaining `prefix_default_language=False` would help future maintainers avoid "fixing" it to `True` and breaking all existing English links.

Everything else looks good — `JavaScriptCatalog` at `/jsi18n/`, `lang="{{ LANGUAGE_CODE }}"` in base.html, JS catalog loading before Alpine. ✅

PR #234, opened 2026-04-03
chore: retire unused chat_message and chat_window components

### LINKED ISSUE #229: chore: retire unused chat_message molecule and chat_window organism
## Summary

Two components are defined but never used in any template (only the style guide). The chat page has diverged and implements its own structures.

### `c-molecules.chat_message`
- Never referenced outside its own file
- Both `chat.html` and the `chat_window` organism hand-roll the chat row structure for Alpine `x-for` loops
- The molecule exists for server-rendered chat messages but that pattern is not used

### `c-organisms.chat_window`
- Only appears in the style guide
- `chat.html` implements its own full chat UI with different Alpine data, different URL patterns, and different message handling
- The organism and page have diverged: organism uses `dynamicMessages` / `chat:stream` URL; page uses `askQuestion` / inline message loop

## Options

1. **Delete both** — they are dead code and maintaining them creates confusion
2. **Refactor `chat.html` to compose the organism** — significant effort, only worth it if the style guide needs a live chat demo
3. **Keep molecule, delete organism** — if server-rendered chat messages become useful later (e.g., initial page load)

Recommend option 1 unless there is a plan to unify.

mitch-mitchel:
Delete chat_message molecule and chat_window organism — both are dead code. chat.html implements its own chat UI with different Alpine data and URL patterns. Update CLAUDE.md component listing to reflect current state.

Closes #229

Test plan: `make test` — no templates reference these components


nadahlberg (approved):
LGTM

PR #275, opened 2026-04-09
chore: make test defaults to quiet output

### LINKED ISSUE #274: chore: make test defaults to quiet output
Flip `make test` to quiet mode by default — suppress tox setup noise and show only failures + summary. Add `make test-v` for verbose output when needed.

Current: `make test` is verbose (3000+ lines of tox setup), `make test-q` is quiet.
Proposed: `make test` is quiet, `make test-v` is verbose.

mitch-mitchel:
`make test` now shows dots + warnings + failures + summary (no per-test lines). `make test-v` for full verbose output. Uses `{posargs}` in tox.ini so Makefile flags cleanly override the default verbosity. Works for both Docker and non-Docker paths.

Closes #274

Test plan: `make test` shows dots/summary, `make test-v` shows full output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help build css clean install migrate test collectstatic lint fmt-check fmt \
+.PHONY: help build css clean install migrate test test-v collectstatic lint fmt-check fmt \
        messages compilemessages \
        docker-build docker-dev docker-prod docker-down docker-logs docker-shell docker-migrate docker-clean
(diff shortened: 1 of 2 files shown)


mitch-mitchel (commented):
LGTM. Clean and focused — `{posargs}` in tox.ini is the right approach, lets `make test` pass quiet flags while `make test-v` falls through to default verbosity.

Issue #261, opened 2026-04-07
feat: resolution state and follow-up UI


mitch-mitchel:
## Goal

Add a resolution/confirmation state after form download and a persistent case history view for authenticated users.

**Current state:** After the action plan page, there's no "what happens next" UI. The `case_clear` endpoint exists but there's no guided wrap-up flow.

## Checklist

- [ ] Confirmation state after form download (next steps, what to expect at court)
- [ ] Case history view for authenticated users (past sessions with case summaries)
- [ ] Deadline reminder scaffold (model + UI placeholder — actual notifications deferred)
- [ ] "Case closed" or "resolved" state on CaseInfo

Part of #172

mitch-mitchel:
Resolution stack complete for the demo (#283, #284, #285, #286 merged).

#287 reframed as a future multi-case enhancement — not needed while auth is email-only demo and users have one case. Tracked independently, blocked on #304.

PR #289, opened 2026-04-15
feat(chat): status-aware endpoints, action item toggle, archive

### LINKED ISSUE #284: feat: status-aware case endpoints, action item toggle, archive-instead-of-delete
## Goal

Make case endpoints status-aware, add toggle endpoints for action items and deadline reminders, and change case_clear from destructive delete to archive.

Part of #261. Depends on #283 (model layer).

## Changes

**`chat/views.py`:**
- `case_get`: filter `status="active"`, include `"status"` in response
- `case_save`: add `status="active"` to `update_or_create` filter, set `chat_session` FK from session
- `case_clear`: archive (`status="archived"`) instead of delete, return `{"archived": True/False}`
- `case_timeline_add`: expand valid_types to include `"resolution"`
- New `case_resolve` (POST): set status="resolved", create resolution timeline event
- New `action_item_toggle` (POST, `<uuid:item_id>`): flip completed, return new state
- New `deadline_reminder_toggle` (POST, `<uuid:deadline_id>`): flip reminder_requested, return new state

**`chat/urls.py`:** Add 3 paths: `case/resolve/`, `case/action-item/<uuid:item_id>/toggle/`, `case/deadline/<uuid:deadline_id>/remind/`

**`chat/agents/litigant_assistant.py`:** Add `status="active"` to both `get_or_create` calls (lines 235, 331)

**Tests (`chat/tests/test_case_views.py`):**
- Update CaseClearTests: assert archived instead of deleted
- Update OwnershipIsolationTests: verify archive behavior
- New: CaseResolveTests, ActionItemToggleTests, DeadlineReminderToggleTests
- New: case_get only returns active cases, case_save creates new active after archive

## Branch

`261-resolution-views` from `261-resolution-models`

~250 lines (may split into views + tests if too large)

mitch-mitchel:
Make case endpoints status-aware, add toggle endpoints for action items and deadline reminders, and change case_clear from destructive delete to archive.

- `case_get`/`case_save` filter by `status="active"`
- `case_clear` archives instead of deleting
- New: `case_resolve`, `action_item_toggle`, `deadline_reminder_toggle`
- Litigant assistant `get_or_create` calls scoped to active cases

Part of #261, closes #284

Test plan: 5 new test classes + updated CaseClearTests and OwnershipIsolationTests for archive behavior

### DIFF
--- a/chat/agents/litigant_assistant.py
+++ b/chat/agents/litigant_assistant.py
@@ -232,7 +232,9 @@ class UpdateCaseFacts(Tool):
                 if session.user
                 else {"session_key": session.session_key}
             )
-            case, _ = CaseInfo.objects.get_or_create(**ownership)
+            case, _ = CaseInfo.objects.get_or_create(
+                status="active", **ownership
+            )
             data = dict(case.data) if case.data else {}
 
             if "case_type" in patch:
@@ -328,7 +330,9 @@ class UpdateActionPlan(Tool):
                 if session.user
                 else {"session_key": session.session_key}
             )
-            case, _ = CaseInfo.objects.get_or_create(**ownership)
+            case, _ = CaseInfo.objects.get_or_create(
+                status="active", **ownership
+            )
             data = dict(case.data) if case.data else {}
 
             # action_items → ActionItemModel (not JSON)
--- a/chat/tests/test_case_views.py
+++ b/chat/tests/test_case_views.py
@@ -11,7 +11,7 @@ import pytest
 from django.contrib.auth import get_user_model
 from django.test import Client, TestCase
 
-from chat.models import CaseInfo, TimelineEvent
+from chat.models import ActionItemModel, CaseInfo, Deadline, TimelineEvent
 
 User = get_user_model()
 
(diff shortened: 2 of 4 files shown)


mitch-mitchel (commented):
## Review

Scope: Backend — converts delete to archive, adds resolve/toggle endpoints, status-aware queries.

Archive over delete is the right call — non-destructive, supports case history. Ownership filtering is consistent across all new endpoints, tests are thorough, rate limiting on all new endpoints.

**Nits / questions:**

1. **`import uuid` inside test methods** (test_case_views.py:369, 393) — should be at the top with the other imports. Works, but inconsistent with the file's style.

2. **Dict comprehension for cross-table ownership filtering:**
   ```python
   **{"case__" + k: v for k, v in ownership.items()}
   ```
   Clever and correct, but if this pattern shows up a third time, consider a helper like `_case_ownership_filter(request)` that returns `{"case__user": user}` / `{"case__session_key": key}` directly.

3. **`case_resolve` doesn't pop `chat_session_id`** — assuming intentional (resolved ≠ cleared), but worth confirming. A user who resolves then starts chatting again would be in the same chat session attached to a resolved case.

4. **No test for resolve + re-save** — there's `test_case_save_after_archive_creates_new_active` but no equivalent for the resolve flow. Can a user save new case data after resolving? The `case_save` filter is `status="active"`, so it would create a new case. Might want a test proving that.

mitch-mitchel:
## Self-review

Clean backend PR. Archive-instead-of-delete is the right call — non-destructive, keeps history, and the lifecycle (active → resolved → archived) is well-defined.

**Highlights:**
- `_ownership_filter` prefix trick in toggle views (`case__` + k) is concise and correct — depends on filter keys matching model fields, which they already do everywhere
- Rate limiting on all new endpoints
- `case_save` with `status="active"` in `update_or_create` correctly creates a new active case after archive — `test_case_save_after_archive_creates_new_active` validates the full lifecycle
- Good test coverage: 5 new test classes, both directions on toggles, 404 for nonexistent IDs

**Note:** `case_clear` moved from `delete()` to `update()`, which means timeline events are preserved (no cascade). The old test that created a timeline event before clearing was removed — correct behavior, just calling it out.

**No blocking issues found.** ✅

Issue #304, opened 2026-04-20
Auth: replace email-only demo auth with robust production auth (SSO, etc.)


mitch-mitchel:
## Context

Current auth is a basic email/password flow (django-allauth) built for demo purposes. It's sufficient for court partner demos but not production.

Interim fixes layered on top of the demo auth (to be folded into the real solution):
- #273 — hide user email in the header (privacy in public spaces); replaced with a generic "You are logged in" indicator

## What's needed for production

- Robust auth provider — SSO (Google, Microsoft, Apple), magic links, and/or passkeys
- Account recovery flows
- Session security (idle timeout, device management)
- Privacy-preserving logged-in indicator (no PII exposed in the UI — see #273)
- Clear differentiation between guest and logged-in experiences (see #121)
- Compliance review: what PII we store, retention, data export/delete

## Open questions

- Which SSO providers do court partners / legal aid orgs already use?
- Do we need to support institutional login (e.g., legal aid org sign-on)?
- What's the minimum viable production auth vs. the full vision?

## Related

- #121 — Logged in vs guest user experience
- #269 — QA: Which sign in do I use? (sign-in consolidation)
- #272 — QA: How does Your Profile use my information?
- #273 — QA: Too much info in logged in display (interim email-hiding fix)

PR #305, opened 2026-04-20
fix(ui): hide user email in header, show generic logged-in state

### LINKED ISSUE #273: QA: April 2026 testing round 1: Too much info in logged in display
**QA environment:** https://litigant-portal.nates.ai

We can't show the user's email for the logged in option. The user could be in a public place (library, court kiosk) and not want people looking over their shoulder to see their email address. Email addresses all give away a lot of personal information (work place, full name, weird nickname you're using as your email address, whatever). 

We can, and should, show that they're logged in, but we shouldn't use their name or email. 

Image

CourtListener doesn't show any information to indicate that you're logged in. We need more than that, but less than this.

mitch-mitchel:
Replace user.email in the header user menu with a translated "You are logged in" label so PII isn't exposed in public spaces (library, court kiosk). The dropdown still has Profile + Sign out, so users can still reach their account. Dropped max-w/truncate since the string is now fixed-length.

This is an interim fix on top of the demo email/password auth — production-grade auth (SSO, etc.) is tracked in #304.

Closes #273
Part of #304

Test plan:
- [ ] Logged in: header shows "You are logged in" (no email) on ≥sm, icon only on mobile
- [ ] Dropdown still opens with Profile + Sign out
- [ ] Logged out: "Sign in" link renders as before

### DIFF
--- a/portal/tests.py
+++ b/portal/tests.py
@@ -168,10 +168,10 @@ class UserMenuAuthenticatedTests(TestCase):
         )
         self.client.login(username="testuser", password="testpass123")
 
-    def test_header_shows_user_email(self):
-        """Header should display user's email when logged in."""
+    def test_header_does_not_expose_user_email(self):
+        """Header must not render the user's email (PII privacy — #273, #304)."""
         response = self.client.get("/")
-        self.assertContains(response, "test@example.com")
+        self.assertNotContains(response, "test@example.com")
 
     def test_header_shows_sign_out_link(self):
         """Header should show 'Sign out' option when logged in."""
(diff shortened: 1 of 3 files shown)


mitch-mitchel (commented):
## Review — request changes

The PR title says hide email "in header," but only the desktop instance was fixed. Mobile still exposes the email.

### Issues (must fix)
- **Mobile account section still shows email.** `templates/cotton/organisms/header.html:113` renders `{{ user.email }}` inside a `sm:hidden` block (mobile-only). This is arguably the *worse* case given the issue's context (library, court kiosk, shoulder-surfing on small screens). Apply the same swap there:
  ```html
  <div class="mb-2 px-1 py-1 text-sm text-greyscale-600">{% trans "You are logged in" %}</div>
  ```
  Drop `truncate` since the string is now fixed-length.

### Suggestions
- The same "You are logged in" string now lives in two templates. Two instances is fine without extraction, but worth noting if a third spot appears.

### Questions
- Should `templates/pages/profile/detail.html:40` (the user's own profile page) also hide the email? Issue #273 text and screenshot both point only at the header, so leaving profile as-is seems right — but confirming with @JessicaFrank in case she meant the whole UI.

### Non-issues (verified)
- `{% trans "You are logged in" %}` as text content is fine — the extract-to-variable rule only applies inside Cotton prop *attributes*.
- No test needed — per CLAUDE.md, template copy changes aren't valid test targets.
- CSP/Alpine surface unchanged.
- a11y: icon + clear status string is an improvement over an email for screen readers.

Issue #310, opened 2026-04-22
Add Adult Name Change to QA site


JessicaFrank:
Request to add Adult Name Change to the LegalFlow on the QA site.

Here is the main ND court's site: https://www.ndcourts.gov/legal-self-help/name-change-adult

Here are [the instructions and the forms](https://www.ndcourts.gov/Media/Default/Legal%20Resources/Legal%20Self%20Help/Name%20Change/Name%20Change%20Adult%20All%20Forms%20%20Instructions.pdf)

The request came from the court. They're interested in seeing what the output of the Litigant Portal would be for this specific issue.

The main name change page has additional legal research resources, additional next steps, and forks in the user's path that we'll want to loop in.

## Scope

Planning and prompt infrastructure (#314 refactor stack) landed 2026-04-23. Demo-time goal: Sandra's happy path clickable end-to-end on QA, with fixture-backed intelligence. Alex (waiver track) is phase-2 stress testing on the same baseline.

## Sub-issues

Build order (bottom-up, each independently demoable):

1. ~~**#330** — chore: align topic slug vocabulary + E2E smoke~~ ✅ merged
2. ~~**#326** — feat(topics): swap expungement → adult-name-change on grid~~ ✅ merged
3. ~~**#327** — feat(routing): deep-link entry at ``/t/{court}/{topic}``~~ ✅ merged
4. **#328** — feat(branding): court name slot in header _(S — in progress)_
5. **#329** — feat(sidebar): pin urgent-priority blockers with warning affordance _(S)_
6. **#335** — feat(prompts): add Official Resources section to Court prompts _(S — demo-scoped court corpus; AI can cite forms PDF + landing pages)_

### Phase 2 (unblocked, not required for Sandra demo)

7. **#331** — feat(prompts): Alex waiver-track phase-2 tightening _(M — uses #312)_

## References

- ``docs/nd-name-change-planning-log.md`` — architecture decisions, scope boundaries, reusable patterns
- ``docs/happy-path-sandra.md`` — standard-track narrative
- ``docs/prompts-as-infra.md`` — Base + Phase + Topic + Court composition contract
- #311 — Sandra user story
- #312 — Alex user story
- #314 — prompt infrastructure parent (landed)

JessicaFrank:
@mitch-mitchel additional FAQs sent in by the ND team 

[...]

Issue #179, opened 2026-03-17
Ground the assistant in court corpus and config on topic entry


mitch-mitchel:
## Summary

The assistant answers from a static prompt today. `LitigantAssistant.generate_system_prompt()` returns a fixed `BASE_PROMPT` and ignores the `thread_id` it is handed, so every conversation gets the same court-neutral instructions no matter which court or topic the user came in through.

This wires the assistant to court config and court corpus, and delivers the ability to load a named court's setup so it can be exercised.

Where this sits in the product shape: today's prod is the **template**. In the full product each mirrored instance of that template is court-specific, determined by the corpus and configs it loads or is built with. QA today is a single shared environment that Jessica and court partners use to test and demo, so there is no way to point it at one court's material. A future QA will be switchable to a court-specific experience — by login, a manual switcher, a URL parameter, something else; **that mechanism is not decided and this issue does not decide it.**

## Why now

#711 names this "the load-bearing item. Nothing else in this section can start in earnest until it exists."

This is foundational product tech, not a feature. The product's claim is that a court-configured assistant grounded in approved material beats a general-purpose model. Nothing demonstrates that today, and nothing can until the assistant reads court material at all.

The failure it prevents is concrete. A litigant who asks the filing fee and acts on a wrong answer is harmed. North Dakota's own two sources disagree on that fee ($80 vs $160, per #711) — exactly the kind of question a model with no corpus answers confidently and wrongly.

## Scope

- The assistant resolves the court and topic for a thread and loads that court's config plus that topic's corpus into its instruction state.
- `generate_system_prompt(thread_id=...)` actually uses `thread_id`.
- A named court's config and corpus can be loaded and exercised end to end, without committing to how a future QA selects between courts.
- Backing data may be fixtures over the current models, or prompt-level content — whatever proves the path. **This does not block on the corpus pipeline (#813) landing.**

Not in scope: conversational track handling (free-form phrasing, follow-ups, objection handling), the eval harness, tool-internal prompts, and authoring real corpus content. Non-fabrication fallback behavior is #744 and pairs with this.

## Definition of done

- [ ] A thread entered from a court and topic produces a system prompt carrying that court's config and that topic's corpus; a thread entered without one is unchanged
- [ ] A named court's config and corpus can be loaded and exercised, from fixtures if the pipeline has not landed. How a future QA switches between courts stays an open question, deliberately
[...]

PR #332, opened 2026-04-23
chore(topics): align grid slug with prompt registry (housing → eviction)

### LINKED ISSUE #330: chore: align topic slug vocabulary (grid → prompts) and E2E smoke
Topic slugs diverge between the UI and the chat layer:

- **Grid (``portal/views.py``):** ``housing``, ``family``, ``small-claims``, ``consumer``, ``expungement``, ``traffic``
- **Prompts (``chat/prompts/topics/``):** ``eviction``, ``adult_name_change``
- **ChatService (``_DEFAULT_JURISDICTION_FOR_TOPIC``):** ``eviction``, ``adult_name_change``

Currently there must be a translation happening somewhere (grid's ``housing`` ends up on the ``eviction`` prompt). Trace it and decide on one canonical vocabulary, then align. If we stick with grid slugs, rename prompt modules + registry keys. If we standardize on prompt slugs, rename grid keys. Either way, pick one.

## Scope

- Find the translation layer (if any) between grid slug and prompt slug
- Decide canonical vocabulary (proposed: prompt slugs — ``eviction``, ``adult_name_change`` — because that's where new topics are added going forward)
- Update the other side to match
- End-to-end smoke: click a topic tile → system prompt composes with the right Topic + Court layers
- Document the slug convention in ``docs/prompts-as-infra.md`` or a new short doc

## Related

- #314 composition work assumed aligned slugs; this is the cleanup pass
- Foundational for the other ND demo sub-issues — land this first

Part of #310

mitch-mitchel:
## Summary

The topic grid used ``housing`` as its slug but the chat prompt registry keys on ``eviction``. ChatService silently dropped the Topic and Court prompt layers whenever users clicked the Housing & Eviction tile, leaving them with BASE + Phase only. This aligns the grid slug with the canonical prompt slug so composition fires end-to-end.

Foundation PR for the #310 ND demo stack. Everything else (deep-link route, court branding, blocker sidebar) assumes aligned slugs.

## Changes

- ``portal/views.py`` — rename ``TOPICS`` key from ``housing`` to ``eviction``
- ``templates/pages/topics/housing.html`` → ``eviction.html`` (git mv, no content change)
- ``portal/tests.py`` — update the 2 assertions that pinned the old slug
- ``chat/tests/test_chat_session_topic.py`` — new regression test ``test_grid_eviction_slug_composes_topic_and_court_layers`` reads the grid's eviction slug from ``TOPICS`` and asserts the composed prompt contains the ``EVICTION`` and ``DUPAGE COUNTY`` anchors
- ``docs/prompts-as-infra.md`` — new "Slug vocabulary" section documenting the one-canonical-slug-per-surface convention

## Test plan

- [x] ``make pre-commit`` locally (lint + tests)
- [x] Regression test fails against the old slug (verified in investigation before the rename)
- [x] No remaining ``housing`` references in production paths (model-field tests retain it as an arbitrary-string fixture; that's fine)
- [ ] Post-merge QA smoke: click Housing & Eviction tile → send a message → verify system prompt contains eviction + DuPage content (manual)

## Follow-ups

- ``_DEFAULT_JURISDICTION_FOR_TOPIC`` in ``chat/services/chat_service.py`` was already keyed on ``eviction`` / ``adult_name_change``, so nothing to change there. When #179 lands, that map gets replaced by request-carried jurisdiction.

Closes #330
Part of #310


mitch-mitchel:
## Self-review

**Scope & shape**
- Small foundation PR, 5 files touched, +49/-5. The template rename is a pure ``git mv`` (content unchanged).
- Single conventional commit; no mid-stack rebases needed since the slug change is surgical.

**Investigation recap (what prompted this PR)**
The grid tile click has been hitting ``?topic=housing`` for weeks, but ``_TOPIC_PROMPTS`` is keyed on ``eviction``. Full trace:

```
Grid click
  → /chat/?topic=housing
  → ChatService(topic="housing")
  → build_system_prompt(topic="housing")
  → _TOPIC_PROMPTS.get("housing") → None   ← silent miss
  → topic + court layers skipped
  → user gets BASE + TRIAGE only
```

No alias existed anywhere (greps confirmed). Users clicking the Housing tile have been getting generic prompts for a while. #310 demo asks would not have worked correctly without this fix.

**What to look at carefully**
1. **Option chosen: canonicalize on prompt slugs** — renamed grid ``housing`` → ``eviction``. Alternative was alias-in-chat-layer; rejected because it institutionalizes two vocabularies. The issue body proposed this direction; just flagging.
2. **Regression test** reads the slug from ``TOPICS`` rather than hardcoding — that way if someone renames the slug again in the future without updating the registry, this test catches it.
3. **Model-field tests retain ``"housing"`` as an arbitrary string** — they exercise the Django ``CharField`` (any string works), not service routing. Intentional.
4. **Doc addition** in ``docs/prompts-as-infra.md`` codifies the convention so future-me (or Nathan) doesn't re-introduce a translation layer.

**Backward compatibility**
- Any external bookmark/link to ``/chat/?topic=housing`` or ``/topics/housing/`` will now 404 / fall back to generic. Demo URLs aren't yet advertised; QA is fresh. Acceptable exposure.
- DB data: any existing ChatSession rows with ``topic="housing"`` will now silently compose without the Topic layer on resume. Non-issue for QA (fresh data); would need a data migration if prod had these. Worth noting for the ND deploy.

**Merge risk:** low. Foundation work; nothing else on main depends on the old slug once this merges. CI on the prior stack was green; this PR adds only new tests + a rename.

PR #336, opened 2026-04-23
feat(branding): court name slot in chat sub-header

mitch-mitchel:
## Summary

Displays the active court's name as a small eyebrow above the topic title in the chat sub-header — so ``/t/nd/adult_name_change/`` shows ``NORTH DAKOTA COURTS`` above ``Adult Name Change``. Institutional trust: the user knows whose self-service surface they're on.

Fourth user-facing piece of the #310 ND demo stack. Builds on #332 (slugs), #333 (grid), #334 (deep-link).

## Where the name comes from

Each court prompt module exports a ``COURT_NAME`` constant co-located with its ``PROMPT``:

```python
# chat/prompts/courts/nd.py
COURT_NAME = "North Dakota Courts"
PROMPT = """..."""
```

New ``get_court_name(court)`` helper in ``chat.prompts`` does a lazy lookup — same pattern as ``_COURT_PROMPTS``. Unknown / empty slug returns ``""`` so callers can treat empty as "no branding" and omit the UI slot entirely.

## Layout stability

Eyebrow only renders when ``court_name`` is set. Topic-only sessions (grid click, no court) look identical to before. No empty slot, no reflow.

- ``.court-branding`` utility in ``src/css/main.css``: uppercase, small, muted — reads as "context label," not content
- Eyebrow sits inside the existing ``topic-entry-header`` flow so it aligns with the topic title, not with the icon column

## Changes

- ``chat/prompts/courts/{nd,dupage_il}.py`` — add ``COURT_NAME`` constant
- ``chat/prompts/__init__.py`` — new ``_COURT_NAMES`` registry + ``_load_court_names()`` + ``get_court_name()`` helper
- ``portal/views.py`` — ``chat_page`` resolves ``court_name`` and passes to template context
- ``templates/pages/chat.html`` — eyebrow line above topic title, conditional on ``court_name``
- ``src/css/main.css`` — new ``.court-branding`` class token
- Tests: 5 on ``get_court_name`` (known, empty, unknown, case-insensitive), 4 on the view/template (context carries name, eyebrow rendered when set, absent when no court)

## Test plan

- [x] ``make pre-commit`` — lint + tests green
- [x] Topic-only session unchanged (no eyebrow, no layout shift)
- [x] Deep-link ``/t/nd/adult_name_change/`` renders the eyebrow
- [x] Deep-link ``/t/dupage_il/eviction/`` renders ``DuPage County Circuit Court`` eyebrow
[...]

PR #334, opened 2026-04-23
feat(routing): deep-link entry at /t/{court}/{topic}

mitch-mitchel:
## Summary

Court-partner deep-link surface. ``/t/nd/adult_name_change/`` validates the pair against the prompt registries, 404s on unknown court or topic, and 302s to ``/chat/?topic=X&court=Y`` so the existing chat page handles the render. Court now flows end-to-end:

```
URL /t/{court}/{topic}/
  → deep_link view (validate, 302)
  → /chat/?topic=X&court=Y
  → chat_page view (validate, render)
  → data-court-slug="..." in DOM
  → Alpine reads + POSTs court alongside topic
  → chat/views.py POST handler
  → ChatService(court=...)
  → build_system_prompt(..., court=...)
```

Third user-facing piece of the #310 ND demo stack. Builds on #332 (slug alignment) and #333 (topic grid). First external entry point the ND court can actually embed.

## Changes

- ``portal/urls.py`` — new ``t/<slug:court>/<slug:topic>/`` pattern
- ``portal/views.py`` — new ``deep_link`` view; ``chat_page`` reads ``?court`` and validates against ``_COURT_PROMPTS``
- ``chat/services/chat_service.py`` — accepts explicit ``court`` kwarg; session still stores jurisdiction (2-letter) via the topic default for backward-compat resume
- ``chat/views.py`` — POST handler reads ``court`` and forwards to ``ChatService``
- ``templates/pages/chat.html`` — ``data-court-slug`` passthrough
- ``static/js/chat.js`` — Alpine reads ``courtSlug`` from dataset, POSTs it alongside ``topic``
- ``portal/tests.py`` — 3 deep-link route tests + 3 chat-page court-param tests
- ``chat/tests/test_chat_session_topic.py`` — 2 ChatService composition tests (court kwarg, court-wins-over-default)

## Test plan

- [x] ``make pre-commit`` locally (lint + tests)
- [x] ``/t/nd/adult_name_change/`` 302s with both params
- [x] ``/t/dupage_il/eviction/`` 302s with both params
- [x] ``/t/xx/adult_name_change/`` 404s (unknown court)
- [x] ``/t/nd/not_a_topic/`` 404s (unknown topic)
- [x] Explicit ``court`` wins over the topic default (e.g. ``eviction`` + ``court=nd`` → composes eviction + ND, not eviction + DuPage)
[...]

PR #333, opened 2026-04-23
feat(topics): swap expungement for adult name change on topic grid

mitch-mitchel:
## Summary

Replaces the Expungement tile on the home-page topic grid with Adult Name Change, matching the canonical prompt slug ``adult_name_change`` from ``chat/prompts/topics/``. The grid slug aligns with the prompt registry so ChatService composes Topic + Court layers end-to-end when users click the tile — regression test guards it.

Second user-facing piece of the #310 ND demo stack (builds on #332's slug alignment foundation).

## Changes

- ``portal/views.py`` — remove ``expungement`` entry from ``TOPICS``; add ``adult_name_change`` entry with three guided ``prompts`` and three ``context_sections`` (standalone vs. divorce-bundled, two tracks, what to know up front)
- New ``templates/pages/topics/adult_name_change.html`` — court-neutral informational content: standalone vs. divorce-bundled, two tracks (standard/waiver), forms, eligibility, process, post-order records cascade, related matters, help
- Delete ``templates/pages/topics/expungement.html``
- Add ``test_grid_name_change_slug_composes_topic_and_court_layers`` — companion to #330's eviction guard. Reads the name-change slug from ``TOPICS`` and asserts ``ADULT NAME CHANGE`` + ``NORTH DAKOTA`` appear in the composed prompt.

## Court-neutral discipline

The topic page stays jurisdiction-free on purpose. ND-specific procedural content (5 forms, $160, Burleigh County, Confidential Information Form attachment rule) lives in ``chat/prompts/courts/nd.py``, not here. When a second jurisdiction adopts adult name change, the topic page survives; only the Court layer changes.

## Test plan

- [x] ``make pre-commit`` locally (lint + tests)
- [x] ``test_all_topics_render_with_correct_context`` picks up the new slug automatically
- [x] New regression test passes — grid slug composes the expected Topic + Court layers
- [x] Manual: navigate to ``/topics/adult_name_change/`` — page renders with sections visible
- [ ] Post-merge QA smoke: click Adult Name Change tile → send a message → verify system prompt contains adult-name-change + ND content

## Known inconsistency (follow-up)

``small-claims`` slug still uses a hyphen. When a small-claims prompt lands, the slug needs to be renamed to ``small_claims`` (Python module names can't have hyphens). Not breaking anything today since no prompt is registered for it. Will file separately if it comes up.

Closes #326
Part of #310

PR #288, opened 2026-04-15
refactor(chat): add resolution model fields and fix Pydantic deprecation

mitch-mitchel:
Add model fields for the resolution state feature and clean up a Pydantic deprecation warning.

**Resolution fields:**
- `CaseInfo.status` (active/resolved/archived) with db_index
- `Deadline.reminder_requested` boolean
- `TimelineEvent` gains `resolution` event type
- `ActionItemModel.to_dict()` now includes `id` and `completed`
- `Deadline.to_dict()` now includes `reminder_requested`

**Pydantic fix:** Replace deprecated `class Config` with `model_config = ConfigDict(extra="allow")` on `Tool` base class — removes the V2 deprecation warning from test output.

Migration is additive (all new fields have defaults).

Part of #261, closes #283

Test plan:
- [x] 175 tests pass, 0 warnings
- [x] New tests: CaseInfo status default/choices, Deadline.to_dict includes reminder_requested, ActionItemModel.to_dict includes id + completed, TimelineEvent accepts resolution type

PR #337, opened 2026-04-23
feat(sidebar): pin urgent-priority blockers with warning affordance

### LINKED ISSUE #329: feat(sidebar): pin urgent-priority blockers with warning affordance
Render ``ActionItem.priority == "urgent"`` items pinned to the top of the sidebar with a ⚠️ icon, per the planning log's "Blockers ≠ traps" pattern.

## Scope

- Sort sidebar action items: urgent first, then normal (preserve within-priority order)
- Visual treatment for urgent items: warning icon prefix, subtle color or border emphasis (use existing tailwind tokens — check ``src/css/main.css`` theme)
- Keep treatment accessible: ``aria-label`` or visually-hidden text explaining the pin, not icon-only meaning
- No new field — the ``priority`` field already exists from #288 (``Literal["urgent", "normal"]``)
- When all urgent items resolve, the pin stays pin-shaped (empty sort bucket collapses cleanly; no layout shift)

## Demo anchor

Sandra's flow pins "Call Burleigh County Clerk about background-check timing" at the top of her sidebar. When she captures the clerk's answer, the blocker resolves and the urgency pin goes away without jarring reflow.

## Test plan

- Sort order: urgent before normal in the rendered list
- Visual regression via screenshot or ``assertContains`` on the warning-icon class
- Accessibility: screen reader announces the priority context

Part of #310

mitch-mitchel:
## Summary

Adds the missing visual + a11y affordance for urgent action items in the sidebar: a warning-triangle icon (heroicon ``exclamation-triangle``) with an ``sr-only`` "Urgent" label, replacing the empty-circle indicator for not-completed urgent items. Completes the "Blockers ≠ traps" pattern from the planning log — urgent items now read as blockers at a glance, not just "red not-done items."

Fifth user-facing piece of the #310 ND demo stack. Builds on #332 (slugs), #333 (grid), #334 (deep-link), #336 (court branding).

## What was already there

Investigation surfaced that most of the machinery was in place from #288 and earlier:

- ``ActionItemModel.Meta.ordering = ["-priority", "title"]`` already sorts urgent first (works because "urgent" > "normal" alphabetically descending)
- ``to_dict()`` already includes ``priority`` in the JSON payload
- ``chat.js`` actionItems getter already computes ``isUrgent`` plus priority-aware color classes (``bg-red-50``, ``border-red-400``, ``text-red-800``, ``text-red-600``)

The gap was purely the warning icon + sr-only label. This PR fills it, plus adds regression guards for the bits that would silently break the affordance if touched.

## Changes

- ``static/js/chat.js`` — new composed flags ``isUrgentOpen`` and ``isNormalOpen`` on each action item. CSP build can't evaluate ``&&`` / ``!`` in templates, so the split has to be pre-computed here
- ``templates/pages/chat.html`` — split the existing ``isNotCompleted`` template block into ``isUrgentOpen`` (warning triangle + visually-hidden "Urgent" label) and ``isNormalOpen`` (existing empty circle)
- ``chat/tests/test_models.py`` — ``to_dict`` priority coverage + ordering regression guard (urgent first, within-priority alphabetical by title)
- ``portal/tests.py`` — template-level assertContains for ``isUrgentOpen`` + ``exclamation-triangle`` markers. Alpine renders client-side, so the test guards that the template markup itself doesn't regress

## Test plan

- [x] ``make pre-commit`` — lint + tests green
- [x] Seed case with mixed priorities via shell → visual check in browser: urgent items pinned at top with warning triangle; normal items below with empty circle
- [x] Toggle an urgent item to completed → red treatment + warning triangle collapse cleanly to the completed-style with green check
- [ ] Post-merge QA smoke: deep link into ``/t/nd/adult_name_change/``, cause the AI to emit an urgent action item, verify the sidebar affordance

## Interaction with completed state

When an urgent item is marked complete, the existing completed-style (``bg-greyscale-50``, greyscale strike-through, green check-circle) takes over automatically — the warning triangle disappears because ``isUrgentOpen`` requires not-completed. No jarring reflow, no lingering red treatment. Matches the planning-log "pin shape stays; empty sort bucket collapses cleanly" intent.

## Known follow-ups

- **Completion-aware sort** — currently a completed urgent item still ranks above normal open items (model sort doesn't consider completion). If the demo surfaces it as confusing, add a tertiary sort key ``completed ASC`` to push completed items to the bottom regardless of priority. Not in this PR.
- **Empty-urgent-bucket state** — when all urgents resolve, the sidebar section flows naturally (no empty header, no placeholder). If product wants an explicit "no active blockers" callout, that's its own UX decision.

Closes #329
Part of #310

### DIFF
--- a/chat/tests/test_models.py
+++ b/chat/tests/test_models.py
@@ -289,3 +289,54 @@ class ActionItemToDictTests(TestCase):
 
         self.assertIn("completed", result)
         self.assertIs(result["completed"], True)
+
+    def test_to_dict_includes_priority(self):
+        """to_dict() must include the priority field (#329 sidebar pin)."""
+        urgent = ActionItemModel.objects.create(
+            case=self.case, title="Call clerk", priority="urgent"
+        )
+        normal = ActionItemModel.objects.create(
+            case=self.case, title="Gather docs", priority="normal"
+        )
+
+        self.assertEqual(urgent.to_dict()["priority"], "urgent")
+        self.assertEqual(normal.to_dict()["priority"], "normal")
+
+
+class ActionItemOrderingTests(TestCase):
+    """Tests for ActionItemModel default ordering (#329 sidebar pin).
+
+    Urgent items must come before normal items so blockers pin to the
+    top of the sidebar. Within a priority bucket, items sort by title.
+    """
+
+    def setUp(self):
+        self.case = CaseInfo.objects.create(data={})
+
+    def test_urgent_items_come_before_normal(self):
+        ActionItemModel.objects.create(
+            case=self.case, title="Gather documents", priority="normal"
+        )
+        ActionItemModel.objects.create(
+            case=self.case, title="Call clerk", priority="urgent"
+        )
+        ActionItemModel.objects.create(
+            case=self.case, title="Review lease", priority="normal"
+        )
+
+        titles = [item.title for item in self.case.action_items.all()]
+
+        self.assertEqual(titles[0], "Call clerk")
+        self.assertEqual(titles[1:], ["Gather documents", "Review lease"])
+
+    def test_within_priority_sorted_by_title(self):
+        ActionItemModel.objects.create(
+            case=self.case, title="Zebra task", priority="urgent"
+        )
+        ActionItemModel.objects.create(
+            case=self.case, title="Alpha task", priority="urgent"
+        )
+
+        titles = [item.title for item in self.case.action_items.all()]
+
+        self.assertEqual(titles, ["Alpha task", "Zebra task"])
--- a/portal/tests.py
+++ b/portal/tests.py
@@ -70,6 +70,18 @@ class ChatPageTests(TestCase):
         response = self.client.get("/chat/")
         self.assertContains(response, "chat-subheader")
 
+    def test_chat_page_renders_urgent_blocker_affordance(self):
+        """Chat page template must include the urgent blocker warning icon (#329).
+
+        The icon is Alpine-rendered conditionally on ``item.isUrgentOpen``,
+        so the raw template markup is what we assert here — if this
+        exclamation-triangle block is removed, the urgent pin affordance
+        regresses silently.
+        """
+        response = self.client.get("/chat/")
+        self.assertContains(response, "isUrgentOpen")
+        self.assertContains(response, "exclamation-triangle")
+
 
 # =============================================================================
 # Auth Template Tests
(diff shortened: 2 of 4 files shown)


mitch-mitchel:
## Self-review

**Scope & shape**
- 4 files, +74/-1. Smallest of the ND demo subs — most of the work was sitting there waiting for this one UI hookup.
- No new dependencies, no new migrations, no new API endpoints.

**What to look at carefully**
1. **Pre-computed ``isUrgentOpen`` / ``isNormalOpen`` flags** (``chat.js``) — these exist because Alpine's CSP build can't evaluate ``&&`` or ``!`` in templates. They duplicate what ``isUrgent`` + ``isCompleted`` already express. Justified by the CSP constraint but worth naming if the actionItems getter feels like it's accumulating flags. Consolidation candidate for a future cleanup — not this PR.
2. **Ordering "just works" via string alphabet** — ``-priority, title`` puts urgent first because ``"urgent" > "normal"`` descending. This is fragile against future priority-string renames (e.g. ``"high"``/``"low"`` would break the sort). The regression test pins the current behavior, but if priority vocabulary changes, the ordering needs to switch to an explicit ``CASE`` or a numeric field.
3. **sr-only "Urgent" label** — placed inside a ``<span class="flex-shrink-0 mt-0.5">`` alongside the icon. Verifies as a single accessible name for the row via VoiceOver. The ``{% trans "Urgent" %}`` wrapping means it localizes if we add i18n later.
4. **Template changes only — no Python view or service changes** — the Alpine computation is the only "code logic" touch, and it's additive (existing consumers of ``isNotCompleted`` / ``isUrgent`` unchanged).
5. **Completed urgent items keep their sort position** — flagged as a follow-up. Clicking an urgent item to complete it leaves it at the top of the list (now with a green check + strike-through) because model ordering doesn't consider ``completed``. If that's visually jarring in demo-testing, file a sort tweak.

**Backward compatibility**
- Non-urgent items unchanged
- Completed items unchanged (they already get the grey/strike-through path, which takes priority over the priority treatment)
- ``isCompleted`` / ``isNotCompleted`` / ``isUrgent`` flags still exported for any other consumer
- No DB migration, no API shape change

**Manual test path** (included in the QA plan I gave you locally)
- Seed ``ActionItemModel`` rows via shell with ``priority="urgent"`` + ``priority="normal"``
- Load chat page attached to that case
- Verify: urgents above normals, warning triangle on urgent-open items, toggle collapses the urgency treatment cleanly

**Merge risk:** low. Single template surface + one JS getter tweak, backed by model-level + template-level tests. Zero impact on non-urgent / completed items.

PR #352, opened 2026-04-30
fix(chat): preserve case state when archive request fails

### LINKED ISSUE #291: fix: clearCaseInfo wipes local state even when server archive fails
## Problem

`clearCaseInfo()` in `chat.js` resets all local state (caseInfo, actionPlan, extractedData, caseTimeline) regardless of whether the server-side archive call succeeds. If a user is 80% through a case and the server call fails (network blip, timeout, 500), their entire case context is wiped locally with no way to recover.

## Current behavior

```js
async clearCaseInfo() {
  // Archive server-side
  try {
    await fetch('/api/chat/case/clear/', { method: 'POST', body: formData })
  } catch (e) {
    console.error('Failed to archive case info:', e)
  }
  // Always clears local state — even on failure
  this.caseInfo = null
  // ...
}
```

## Expected behavior

Only reset local state after confirmed server-side archive. On failure, show an error (toast?) and keep the case intact.

## Context

Found during self-review of PR #290. Same pattern existed before (delete instead of archive), but the stakes are higher now that cases have richer state (action items, deadlines, timeline events).

Part of #261

mitch-mitchel:
## Summary

- `clearCaseInfo()` previously reset all local case state (`caseInfo`, `actionPlan`, `extractedData`, `caseTimeline`) regardless of whether the server-side archive call succeeded — a network blip or 5xx silently wiped the user's case context with no recovery path.
- Now the local reset only runs after the server confirms (`response.ok`). On failure, we set `caseInfoError`, surface a `c-atoms.alert variant="danger"` banner in the case sidebar (with `aria-live="polite"`), and keep the case intact so the user can retry.
- Two-file diff: `static/js/chat.js` (state field + rewritten method) and `templates/pages/chat.html` (banner inside the case sidebar slot).

## Test plan

- [ ] Successful clear: case info wipes from sidebar, no banner shown.
- [ ] Stop django container or block `/api/chat/case/clear/` (e.g., return 500), click **Clear** → banner appears, case info stays intact in sidebar.
- [ ] Retry after a transient failure → next successful clear wipes state and clears the banner.
- [ ] Screen-reader: `aria-live="polite"` announces the banner text on appearance.

## Coverage gap

The repo has no JS test harness, so the early-return path is not under automated coverage. The server-side `case_clear` view is unchanged and continues to be covered by `chat/tests/test_case_views.py::CaseClearTests` (postgres-marked). Worth a follow-up issue to either (a) introduce a JS test harness, or (b) extract enough state into the server response that the JS branch is exercisable via Django integration tests.

Closes #291

### DIFF
--- a/static/js/chat.js
+++ b/static/js/chat.js
@@ -441,6 +441,7 @@ function createHomePage() {
     showConfirmation: false,
     _documentContextSent: false,
     _isClearingChat: false,
+    _isClearingCase: false,
 
     // --- Action plan state ---
     actionPlan: null, // { action_items: [], spotted_issues: [], resources: [] }
@@ -1247,6 +1248,12 @@ function createHomePage() {
       // Archive server-side (non-destructive). Only reset local state after
       // the server confirms; otherwise a network blip or 5xx would silently
       // wipe the user's case context with no recovery path.
+      // Guard against rapid double-click — without it, a failing fetch
+      // resolving after a succeeding one would re-set the error banner over
+      // an already-cleared state.
+      if (this._isClearingCase) return
+      this._isClearingCase = true
+
       const formData = new FormData()
       formData.append('csrfmiddlewaretoken', chatUtils.getCsrfToken())
 
@@ -1265,6 +1272,7 @@ function createHomePage() {
         this.caseInfoError = gettext(
           "Couldn't clear your case — please try again."
         )
+        this._isClearingCase = false
         return
       }
 
@@ -1276,6 +1284,7 @@ function createHomePage() {
       this.extractedData = null
       this.caseTimeline = []
       this._documentContextSent = false
+      this._isClearingCase = false
     },
 
     // =========================================================================
--- a/static/js/chat.js
+++ b/static/js/chat.js
@@ -437,6 +437,7 @@ function createHomePage() {
     // --- Case management state ---
     caseInfo: null,
     caseStatus: '',
+    caseInfoError: null,
     showConfirmation: false,
     _documentContextSent: false,
     _isClearingChat: false,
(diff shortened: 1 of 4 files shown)


mitch-mitchel:
## Self-review

**Approve** — core change is correct, lean, and lands the right invariant (no local wipe without server confirmation). A few minor suggestions, fixing two below before merge.

### Issues (must fix)

None.

### Suggestions (fixing inline)

- **Race condition on rapid double-click.** No debounce — if the second fetch resolves OK before the first resolves with `!ok`, local state wipes from the second, then the first overwrites `caseInfoError` to set the banner. Leaves an empty sidebar showing "Couldn't clear your case." Mirroring the existing `_isClearingChat` flag pattern would make `clearCaseInfo` idempotent. **Fixing.**
- **Redundant `aria-live="polite"` on the wrapper.** `c-atoms.alert` already hardcodes `role="alert"` (implies `aria-live="assertive"`). Screen readers follow the inner role, so the outer `polite` doesn't do anything — and if it did, it would conflict. **Fixing.**

### Suggestions (out of scope, follow-up issue)

- **Sibling anti-pattern in `clearDemo()`** (`static/js/components.js:68`) — same shape: silent fetch, no error handling, then `location.reload()`. Smaller blast radius (demo reset is destructive by design) but identical failure mode. Will file as a tech-debt follow-up.

### Questions (taste calls, not changing)

- **Message tone.** "Couldn't clear your case — please try again." is slightly clinical; could soften to "We couldn't clear your case right now — please try again." Leaving as-is.
- **Manual verification still pending.** Test plan items 2–4 require running locally with the endpoint blocked. No JS test coverage to catch a regression here.

### Files reviewed

- `static/js/chat.js` — state field placement near `caseInfo`/`caseStatus` is correct (case-management state cluster); method body is straightforward; CSP-build constraints respected (no expressions, just property/method refs).
- `templates/pages/chat.html` — banner placement above both empty-state and case-info-display means it's visible regardless of which state is active; `x-cloak` correctly paired with `x-show`; uses existing `c-atoms.alert` rather than inventing a new component.

mitch-mitchel:
## Fixes applied

Self-review suggestions landed in `d8cf180`:

- ✅ **Race-debounce** — `_isClearingCase` flag added to state; `clearCaseInfo` early-returns if already in flight, clears the flag on both success and failure paths.
- ✅ **Redundant `aria-live="polite"`** — dropped from the wrapper; comment notes the alert atom owns `role="alert"`.

### Still outstanding (not blocking merge)

- **`clearDemo` sibling anti-pattern** (`static/js/components.js:68`) — same shape, smaller blast radius. Will file as a tech-debt follow-up issue rather than expand this PR's scope.
- **Manual negative-path verification** — test plan items 2–4 still pending; no JS test harness to catch a regression.

Ready for merge.

Issue #356, opened 2026-04-30
Define slug naming convention for states and courts


mitch-mitchel:
## Context

[#338](https://github.com/freelawproject/litigant-portal/issues/338) just shipped canonical hyphenated court slugs (`north-dakota`, `dupage-il`). That worked because each state currently has exactly one registered court, but the convention won't scale once we have:

- Multiple courts per state (ND has different levels — state supreme, district, county; AZ partnership lands with ≥1 court).
- State-level deployments (e.g., `portal.arizona.gov`).
- Cross-court overlays (state-wide rules + court-specific tweaks).

[#349](https://github.com/freelawproject/litigant-portal/issues/349) (AZ proposal) is the forcing function — first AZ meeting **Wed May 6, 5pm ET**. Convention should land before then so AZ's first court onboards under the right slug.

## Today's state

`_COURT_PROMPTS` registry keys (post-#338):

- `north-dakota` — really "ND state-level courts as a system"
- `dupage-il` — DuPage County, IL (state-suffixed, county-prefixed)

Both work as URL-facing slugs (`/t/{court}/{topic}/`). Neither generalizes:

- What slug does ND's *district* court get when it onboards? Where does that put `north-dakota`?
- `dupage-il` has the state at the end; AZ's analogous county reads naturally as `arizona-maricopa`. Inconsistent ordering.

### State / portal slug (one per state partnership)

`arizona`, `north-dakota`, `illinois` — the overall partnership; doubles as a portal subdomain target (`portal.arizona.gov`); resolves to a default court for unspecified deep-links.

### Court slug (one per registered court within a state)

State-first, hyphen-separated:

- `arizona-supreme`
- `arizona-maricopa-superior`
- `arizona-tucson-municipal`
- `north-dakota-supreme`
- `north-dakota-burleigh-county`
- `illinois-dupage-county`

### Court-type vocabulary (open list, grows with partners)

- `-supreme` — state supreme court
- `-{county}-county`
- `-{city}-municipal`
- `-{circuit-name}-circuit`
- `-{name}-appeals`

### State slug as default-court alias

`/t/north-dakota/adult_name_change/` resolves to ND's primary registered court (today: the only one; tomorrow: configurable per state). Solves "what URL do I share if I don't know the specific court?" without adding routing complexity. State-level partnership behavior keeps working as multi-court grows.

### Flat namespace (recommended)

[...]

PR #357, opened 2026-04-30
fix(routing): canonical hyphenated court slugs (north-dakota, dupage-il)

### LINKED ISSUE #338: Deep-link route 404s on human-readable court slugs (e.g. /t/north-dakota/)
## Problem

`/t/{court}/{topic}/` (added in #334) validates `court` against `_COURT_PROMPTS` and raises `Http404` on miss. The registry currently keys ND as `nd`, so a human-readable share URL like:

    https://qa.litigantportal.com/t/north-dakota/adult_name_change/

returns 404, while `/t/nd/adult_name_change/` works.

Same shape as the topic slug mismatch #330 caught — but one layer up (court instead of topic).

## Why it matters

Deep-link URLs are a demo-facing surface: partners are the most likely to share/bookmark `/t/<court>/<topic>/`. The intuitive court slug is the human name, not the internal abbreviation. No grid or JS currently generates this URL, so it isn't surfaced today — but it will 404 the moment anyone tries the obvious share URL.

## Proposed fix options

- **(a) Register courts under human slugs** (`north-dakota` instead of `nd`, `dupage-il` stays). Matches the "canonical slug per surface" rule in `docs/prompts-as-infra.md` → one slug, used everywhere.
- **(b) Add an alias map in `deep_link`** — accept either `north-dakota` or `nd`, normalize to the canonical before composition. More forgiving, but now there are two sources of truth.
- **(c) Move resolution into a helper** — `resolve_court_slug(raw) -> canonical | None` — so the grid, deep-link, and future surfaces all share the same logic.

Leaning (a) for consistency with the slug-vocabulary convention. (c) if we need aliasing for other reasons later.

## Regression guard

Extend the test pattern from #332 to courts: read the court slug from the UI source (grid data attrs / deep-link doc) and assert round-trip through `_COURT_PROMPTS`. Any future (court, topic) pair added should be caught the same way.

## Repro

    curl -sI https://qa.litigantportal.com/t/north-dakota/adult_name_change/
    # HTTP/2 404

    curl -sI https://qa.litigantportal.com/t/nd/adult_name_change/
    # likely HTTP/2 302 → /chat/?topic=adult_name_change&court=nd

Found during post-merge drift audit on 2026-04-24.

mitch-mitchel:
> [!IMPORTANT]
> **Draft pending #356** — slug naming convention is under review by @JessicaFrank. If the convention shifts (e.g., `dupage-il` → `illinois-dupage-county`, or `north-dakota` reserved as state-alias with the actual court keyed as `north-dakota-supreme`), the renames land as additional commits on this branch rather than a follow-up PR. Promotes to ready-for-review once #356 is locked.

## Summary

The deep-link route `/t/{court}/{topic}/` validated `court` against the prompt registry, but the registry keyed ND as `nd` and DuPage/IL as `dupage_il`. Partner-facing share URLs like `/t/north-dakota/adult_name_change/` — the obvious shape someone would type or bookmark — returned 404.

Re-keys `_COURT_PROMPTS`, `_COURT_NAMES`, and `_JURISDICTION_TO_COURT` values to canonical hyphenated slugs (`dupage-il`, `north-dakota`). Module filenames stay underscored (Python identifiers can't have hyphens); the **registry key — the URL-facing surface — decouples from the module name**. Pattern proposed for now: `lowercase-state-name` for state-level, `lowercase-county-state` for county-level. Pending #356 review by Jessica before locking.

## What changed

- `chat/prompts/__init__.py` — registry key rename, jurisdiction-map values, docstring.
- `portal/tests.py` — deep-link tests use canonical slugs; new `test_legacy_underscore_court_slug_returns_404` regression guard; new `test_every_registered_court_resolves_via_deep_link` round-trip in the spirit of #332.
- `chat/tests/test_prompts.py` — court arg in tests use canonical slugs (jurisdiction args stay as 2-letter state codes — that key didn't change).
- `chat/tests/test_chat_session_topic.py` — same court-arg updates; comment clarifies that session.jurisdiction is independent of the court slug.
- `docs/nd-name-change-planning-log.md` — example URL.

## Test plan

- [ ] Existing deep-link tests pass with canonical slugs (`/t/north-dakota/...`, `/t/dupage-il/...`).
- [ ] New legacy-slug 404 test confirms `/t/nd/...` and `/t/dupage_il/...` no longer resolve.
- [ ] New round-trip test walks every key in `_COURT_PROMPTS` and asserts each resolves — catches a future court being registered under a non-canonical slug.
- [ ] Court-name lookups still work (`get_court_name("north-dakota")` → "North Dakota Courts").

## Coverage / safety notes

- `chat_service.py`'s `_DEFAULT_JURISDICTION_FOR_TOPIC` map was not changed — its keys are 2-letter jurisdictions, not court slugs. The jurisdiction → court resolution happens in `_JURISDICTION_TO_COURT`, whose values were updated.
- DuPage/IL is a demo starter partner with no live URLs being tracked; the `dupage_il` → `dupage-il` rename is partner-side risk-free.
- ND is a real partner but no UI currently surfaces the URL in shareable form; the rename gets ahead of the obvious-share-URL bug rather than fixing one users have hit.

Closes #338


mitch-mitchel:
Self-review note from #369: after this lands, the per-court directories `chat/prompts/courts/dupage_il/` and `chat/prompts/courts/nd/` will need to migrate to whatever canonical slug shape #356 settles on (e.g. `dupage-il`).

The PR that migrated prompts to a per-slug directory layout (#369) preserved the existing underscored slugs; renaming dirs + updating any registry callsites is a clean follow-up once the slug convention locks.

Issue #420, opened 2026-05-18
Move dockerfile, entrypoint, and install-tailwind to docker/django


nadahlberg:
Follows the FLP pattern #400

PR #425, opened 2026-05-18
Move django scripts to docker dir

nadahlberg:
Move docker related scripts to their appropriate folders in the `docker` dir.


nadahlberg:
Fixes #420

Issue #497, opened 2026-06-09
Integrate S3 and setup django storages


nadahlberg:
Setup Django storages for private and public S3 objects.

PR #516, opened 2026-06-11
Add storage

nadahlberg:
This PR sets up object storage for the app. A few things:

- Drops whitenoise, serves static files from S3 bucket
- Adds S3 storage for public and private buckets
- Uses filesystem storage locally


nadahlberg:
Fixes #497

PR #552, opened 2026-06-18
Partner/QA ops: reboot durability + one-command update tool

### LINKED ISSUE #548: fix(deploy): QA/prod containers don't auto-restart after reboot
## What happened

A droplet reboot (today's QA RAM resize) left the LP stack **down** — `django-prod`, `caddy-prod`, `postgres-prod`, `redis-prod` did not come back, and the site was unreachable until a manual `docker compose --profile prod up -d`.

## Root cause

None of the services in `docker-compose.yml` declare a `restart:` policy. Docker starts on boot (confirmed — the daemon was up after the reboot), but with no restart policy it has no instruction to bring the containers back. Any reboot, crash, or OOM leaves QA/prod dark until someone SSHes in.

## Fix

Add `restart: unless-stopped` to the four **prod** services (`postgres-prod`, `redis-prod`, `django-prod`, `caddy-prod`). Dev services are intentionally left alone — a laptop shouldn't auto-resurrect the full stack on every boot.

- `unless-stopped` (not `always`): a deliberate `docker compose down` stays down; only reboots/crashes trigger restart.
- The docassemble compose already sets `unless-stopped`, so it's covered once started.
- No CD change: `docker compose --profile prod up -d` recreates containers on the next deploy, applying the policy.

## Definition of done

- Four prod services carry `restart: unless-stopped`.
- Running QA box patched live (`docker update --restart unless-stopped`) so it's reboot-safe before the next deploy.
- Verified: `systemctl is-enabled docker` = enabled (belt-and-suspenders).

### LINKED ISSUE #551: feat(ops): one-command update script for partner-hosted boxes
## What

A single `scripts/update.sh` (+ `make update`) so a court-partner admin can update on demand without remembering the multi-file compose incantation. The script encapsulates the whole sequence:

1. `git pull` (latest code: compose, Caddy, scripts) — skippable with `--no-code`.
2. `docker compose <the right -f files> pull` + `up -d` to recreate the selected service(s) from new images.
3. Service-specific: Django → migrate + collectstatic; docassemble → wait for boot.
4. **Health/validation summary** — hit each service's health surface, print a clear per-service ✓/✗.

Flags: `--all` (default) / `--django` / `--docassemble` (update one or both), `--no-code`, `--check-only` (health only, no pull). Auto-detects whether docassemble is hosted (includes its override file only when its container is running or it's explicitly selected).

## Why

Partner easy-mode: the real deploy command is `docker compose -f docker-compose.yml -f docker-compose.docassemble.qa.yml --profile prod ...` — a lot of syntax an admin runs rarely and won't remember. One command they can. **Opt-in by design** — a normal restart/reboot never auto-updates (that's the #548 restart policy on existing images); this is the explicit path to take new fixes.

## Definition of done

- `scripts/update.sh` + `make update` with the flags above.
- Updates one service without touching the other.
- Ends with a health/validation summary.
- Documented for partners (runbook section).

Pairs with #548 (reboot durability is the no-auto-update half; this is the on-demand-update half).

mitch-mitchel:
Two halves of the same partner/QA ops lifecycle:

1. **Reboot durability** — `restart: unless-stopped` on the four prod services, so a droplet reboot brings the LP stack back on its own. Today nothing restarts the containers after a reboot (a resize today left QA down until a manual `compose up`). Dev profile left untouched.
2. **One-command updates** — `scripts/update.sh` (+ `make update` / `make health`): pull code + images → recreate → health summary, or any single step (`code`/`pull`/`restart`/`health`) on its own, with `--django` / `--docassemble` / `--all`. Opt-in by design — a reboot never auto-updates (reuses existing images); this is the explicit path to new fixes.

Closes #548. Closes #551.

Test plan: restart policy applies live via `docker update --restart unless-stopped` (and on next deploy); update script is syntax-checked and `--help` renders — first real run is the docassemble update on QA.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: help build css clean install migrate test test-v collectstatic lint fmt-check fmt \
        file-issue messages compilemessages docker-build docker-dev docker-prod docker-rebuild \
 	   docker-down docker-logs docker-shell docker-migrate docker-clean \
-	   docassemble-up docassemble-down
+	   docassemble-up docassemble-down update health
 
 help: ## Show this help message
 	@echo "Available commands:"
@@ -113,3 +113,9 @@ docassemble-up: ## Start the local-dev docassemble bench (http://localhost:8100)
 
 docassemble-down: ## Stop the local-dev docassemble bench
 	docker compose -f docker-compose.docassemble.yml down
+
+update: ## Update a hosted box: pull code+images, recreate, health (ARGS=... e.g. --docassemble)
+	./scripts/update.sh $(ARGS)
+
+health: ## Health/validation summary for a hosted box (no changes)
+	./scripts/update.sh health $(ARGS)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,6 +38,10 @@ services:
   postgres-prod:
     image: pgvector/pgvector:pg17
     profiles: ['prod']
+    # Reboot/crash durability for the always-on QA/prod box. unless-stopped (not
+    # always) so a deliberate `compose down` stays down. Dev services omit this
+    # on purpose — a laptop shouldn't resurrect the full stack on every boot.
+    restart: unless-stopped
     environment:
       - POSTGRES_DB=litigant_portal
       - POSTGRES_USER=postgres
@@ -70,6 +74,7 @@ services:
   redis-prod:
     image: redis:7-alpine
     profiles: ['prod']
+    restart: unless-stopped
     command: redis-server --save "" --appendonly no
     healthcheck:
       test: ['CMD', 'redis-cli', 'ping']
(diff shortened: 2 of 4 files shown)


mitch-mitchel:
Both findings addressed in f85c60c:
- **Recovery silent-skip:** `--all` now detects docassemble via `docker ps -a` (container exists in *any* state), so a stopped/crashed docassemble is still updated + health-checked under the default flow — the recovery case. I used `docker ps -a` rather than the suggested `[[ -f $DA_OVERRIDE ]]`: that override file is committed to the repo (via #554), so it's present on every box and would make `--all` try docassemble even on a Django-only box.
- **`--help` leak:** now prints the leading comment block via awk, stopping at the first non-comment line — no leaked shell.

On the dead link to `docs/docassemble-qa-hosting.md`: that doc is real but lives in #554, so it resolves on `main` as long as #554 merges with or before this PR (merge-order note, not a code change).

Issue #549, opened 2026-06-17
docassemble demo UI/UX: branding + style parity + polish


mitch-mitchel:
## What

Make the docassemble interview surface feel like Litigant Portal for court-partner demos — branding + style parity + remove dead/stock affordances. Uses LP's design system as the **source** for a lightweight match (palette / type / logo from `litigant_portal/app/src/main.css` + `static/images/logo.svg`) — **not** a full atomic-design port into docassemble (deferred).

### Checklist
- [ ] **Download-only end screen** — `allow emailing: False` on both interviews. The email option isn't wired (no SMTP) so it's a dead button; no SMS option exists. (folded in from the original #549 scope)
- [ ] **LP palette + fonts** applied via custom CSS — primary `#b5362d`, greyscale ramp, Inter.
- [ ] **Logo + brand name** in the navbar; drop the `docassemble` identity + favicon.
- [ ] **Hide the stock face** — no sample/demo interviews, no Playground/admin or login wall for an anonymous litigant; 'Fill out your forms' lands straight in the interview.
- [ ] **Versioned + reproducible** — CSS / logo / config live in the repo and mount into the container, not hand-set in the web UI.

## Why

A raw `docassemble`-branded, login-walled page with a dead 'email a copy' button reads as a third-party bolt-on in front of court partners. Parity makes the Topic Flow → document-assembly handoff feel like one product.

Part of #531. Develop/test on the local bench first, then deploy to QA.

Issue #441, opened 2026-05-27
Handoff rendering — decision record + file-artifact sub-issue tracker


mitch-mitchel:
### Decision record

Handoffs are a **render-time / authored concern, not a data entity.** The corpus schema models no Handoff type: contact fields (`phone`, `email`, and the like) are neutral strings, and whether/how to surface one as an affordance is decided elsewhere. This issue exists to record that decision and to track the work that *is* real, so it isn't re-litigated per handoff.

**Validity is guaranteed upstream, not at click time.** The engine and UI are trusting consumers of a validated corpus contract (open-contracts principle, #472/#475). A malformed contact or date is a corpus-authoring / load-time failure (Pydantic + loader cross-refs + startup check), never something the render view or template re-checks when a user taps a link. This closes the "validate the Contact before generating the .vcf" question — by the time a litigant clicks, the data is already known-good.

The two kinds of handoff resolve in two different places:

**Native URL-scheme links (`tel:` / `sms:` / `mailto:`) — no engine code.**

- *Which* scheme to use, and whether to surface a number/email at all, is a content decision → the future authoring tool / dashboard milestone (#533).
- The native behavior (dial / text / compose mail) is the browser/OS's job once a standard href is emitted. We build nothing for it.
- `Contact.phone` stays a neutral string (a `fax` would be too); the engine never infers call-vs-text from field presence.

**Declared file artifacts (`.vcf` / `.ics` / packet / summary) — real engine code, shipped as sub-issues.**

These are genuine engine contracts — "given this data shape, produce this file" — independent of authoring. Each ships as its own small sub-issue:

- [x] `.vcf` — Contact → vCard (pure function of a Contact, no user input) — #473
- [x] `.ics` — Deadline + gathered date → calendar event (v1, this milestone) — #504
- [ ] email delivery — packet / progress (**v2 — deferred**, ties to #184/#177)

Cross-cutting constraint for all file artifacts: progressive enhancement (works without JS), CSP compliance, and a download/delivery path — designed before the first generator lands.

Part of #389.

mitch-mitchel:
Design reference from the LawDroid review (#471): their deadline model is a useful shape to borrow for the deadline work here — it tracks the deadline's type, its current status, when reminders should go out, and how a deadline flagged early carries through to the later steps. Noting it so it's on hand when we build the deadline + `.ics` piece.

mitch-mitchel:
[...]

PR #553, opened 2026-06-18
feat(docassemble): LP branding + download-only demo interviews

mitch-mitchel:
Make the ND name-change interviews read as Litigant Portal for court-partner demos, and drop the dead email affordance.

- `lp-branding.css` maps LP's palette + type onto docassemble's Bootstrap (parity via CSS variables, not a design-system port); wired into both interviews via `features: css`. `lp-logo.svg` staged.
- `allow emailing: False` → download-only end screen (email delivery isn't wired; deferred to v2 #441; no SMS).
- README documents the bench apply steps + the global brand/anonymous Configuration that remains.

Part of #549 — the server-side brand/anonymous config + packaging for a clean anonymous URL are bench follow-ups on the tracker.

Test plan: interview YAML validated (8 docs each; `features` + `allow emailing` parse); apply CSS+logo on the local bench and eyeball.


mitch-mitchel:
Verified against docassemble source (`parse.py:3295` reads `allow emailing` from the top-level question dict; line 2174 lists it as a question-level key), so this was a real bug — nested under `attachment:` it was silently ignored and the email button still rendered. **Fixed in 5fd8df9:** moved `allow emailing: False` to question level (sibling of `attachment:`) in both interviews, with a comment so it doesn't regress.

Issue #550, opened 2026-06-17
feat(infra): docassemble on EKS for QA — live single container


mitch-mitchel:
## What

Run docassemble in the `qa-litigant` namespace so `/interview/` serves on `qa.litigantportal.com`, and devs have a shared instance to author interviews on.

The all-in-one `jhpyle/docassemble` image, in live mode. Not a locked-down service: devs log in, edit interviews in place, upload packages, and test PDF field mappings against the running instance.

- Mounted under the `/interview/` path on the QA ingress, not a subdomain (one-CNAME principle).
- Persistent volumes on `/usr/share/docassemble/backup` and `/files`. The container backs itself up there and restores on boot; without them a recreate wipes every interview and resets the admin login to a public default (#701).
- Full `stop_grace_period`, 600s in our compose. A pod killed before the backup finishes comes back on stale state.
- `USEHTTPS=false`, `BEHINDHTTPSLOADBALANCER=true`, verified from the image source. The ingress terminates TLS; docassemble serves plain HTTP and trusts the proxy.
- No public port beyond the ingress path.

## Why

QA has no docassemble today. Neither `deploy.yml` nor `qa-deploy.yml` creates one, and the only instance we deploy ourselves is on the DigitalOcean box, which is `dev` now. Until this lands, `/interview/` 404s on QA, #908 has nothing to configure, and #585's form-completion steps can't be verified.

Scope is the QA/dev/demo iteration cycle. Court-instance hosting is a separate decision (#888) and nothing here constrains it. The all-in-one is chosen precisely because this is dev tooling.

The original scope here was a Caddy-override serving `interview.qa.litigantportal.com` on the DigitalOcean box. That shape is superseded: QA is EKS, and the box is dev.

## Definition of done

- [x] docassemble serving at `qa.litigantportal.com/interview/` over HTTPS
- [x] Interviews, config and accounts survive a redeploy of `litigant-web`
- [x] Admin plus a shared dev account provisioned, credential handling written down
- [x] `nd-name-change` interviews installed as a real package, not a Playground copy (#908)
- [x] Documented in `docs/docassemble.md`

Pairs with #908 (config wiring) and #885 (the access gate). Verification is #587.

mitch-mitchel:
**Re-scoped 2026-08-12: AWS, not the DigitalOcean QA box.**

[...]

PR #554, opened 2026-06-18
feat(infra): docassemble at /interview/ over HTTPS (path-routed)

mitch-mitchel:
Serve docassemble at `https://<lp-host>/interview/` so the Topic Flow link-out is a clean, partner-friendly URL — routed **under the existing hostname** (one CNAME for the whole portal), not a subdomain.

- QA-only compose override adds docassemble to the LP network (no public port); Caddy `handle /interview/*` → docassemble; `POSTURLROOT=/interview/`; behind-proxy env (`USEHTTPS=false`, `BEHINDHTTPSLOADBALANCER=true`) verified against jhpyle/docassemble source.
- Prod untouched — override-only `conf.d` mount + empty-glob import.
- Codifies the **'low-friction for partners / one hostname, path-routed services'** guiding principle in ARCHITECTURE.md.
- Runbook: `docs/docassemble-qa-hosting.md`.

Part of #550 — closes once deployed + verified on QA (set `DA_DOMAIN`, deploy the override, test the `/interview/` WebSocket).

Test plan: compose merge verified with `docker compose config` (service + behind-proxy env + caddy `conf.d` mount); env names verified against source. End-to-end is the QA deploy.

### DIFF
--- a/docker-compose.docassemble.qa.yml
+++ b/docker-compose.docassemble.qa.yml
@@ -10,8 +10,9 @@
 # routes /interview/* to docassemble under the existing LP hostname. Base-compose
 # -only deploys (prod) see none of this.
 #
-# Prereqs: DA_DOMAIN in .env = the existing LP host (e.g. qa.litigantportal.com) —
-# NO new DNS record, the interview lives at <host>/interview/. QA at >=4GB RAM
+# Prereqs: DA_HOSTNAME in .env = the existing LP host as a BARE hostname, e.g.
+# `qa.litigantportal.com` — NO scheme (unlike DOMAIN, which is `https://...`),
+# and no new DNS record; the interview lives at <host>/interview/. QA at >=4GB RAM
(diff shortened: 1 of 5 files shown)


mitch-mitchel:
All three findings addressed in 5d0159d:
- **`DA_DOMAIN` scheme footgun:** renamed to `DA_HOSTNAME` (mirrors docassemble's own `DAHOSTNAME`) and documented that it's a **bare host**, unlike scheme-prefixed `DOMAIN`. (The `${DOMAIN#https://}` option doesn't work in Compose interpolation — no bash expansion — so rename + explicit note.)
- **Bare `/interview`:** added `redir /interview /interview/ 308` before the handle block.
- **Troubleshooting bullet:** corrected — Caddy uses `DOMAIN`, not `DA_HOSTNAME`, so an unset value surfaces as docassemble building wrong URLs, not Caddy failing to start.

PR #573, opened 2026-06-26
chore(ci): Adds Docker Hub build-and-push workflow

ERosendo:
### Summary

Adds a CI workflow that builds the litigant-portal Docker image and publishes it to Docker Hub on every merge to main, along with the required Makefile targets.

This is the initial version and only handles publishing the Docker image.

### Runs alongside cd.yml

This workflow is independent of cd.yml (GHCR build + QA deployment). Each uses its own concurrency group, so they can run in parallel without blocking one another.

One consequence is that every merge to main now builds the image twice: once for Docker Hub and once for GHCR.


chaco-fl (approved):
LGTM

PR #583, opened 2026-06-30
feat(chat): implement hidden messages

### LINKED ISSUE #577: Implement hidden messages
Add a utility for injecting a hidden user message and enforce these actually not being sent to the frontend.

nadahlberg:
Fixes #577

Adds hidden-message support to the chat_v2 engine: messages flagged hidden remain in the history sent to the LLM but are excluded from everything sent to the frontend (thread reloads and sidebar snippets) via a new chat_message_list_visible selector. Includes a chat_message_inject_hidden service helper for seeding hidden context into a thread, and updates AGENT_DEV_GUIDE.md to document the feature.

### DIFF
--- a/docs/AGENT_DEV_GUIDE.md
+++ b/docs/AGENT_DEV_GUIDE.md
@@ -375,44 +375,62 @@ engine then derives two views explicitly:
 ## Conversation history: hidden & compaction messages
 
 Two message capabilities extend the history model without changing how agents
-are authored. Both are pure engine behavior driven by per-message flags — agents
-and tools opt in, the engine does the rest.
+are authored. Both are engine behavior driven by per-message flags — agents and
+tools opt in, the engine does the rest.
 
 ### Hidden messages
 
-A **hidden message** exists in the conversation history that the model sees, but
-is **never returned to the frontend**. It's flagged `hidden=True` on the message.
+A **hidden message** lives in the history the model sees but is **never returned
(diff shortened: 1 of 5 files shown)


mitch-mitchel:
Reviewed — correct and clean. The split holds: `chat_stream` keeps `chat_message_list` (all) so hidden messages still reach the LLM, while both frontend projections (`thread_render_items` + the `thread_list` snippet) switch to `_visible`. `default=False` + the migration keep existing messages visible, and the "inject doesn't bump `updated_at`" note is accurate (no signal; the thread is only bumped explicitly).

One ask before merge: it's a behavioral contract with no tests. Three would pin it —

1. `chat_message_inject_hidden` stores `hidden=True` and leaves `thread.updated_at` unchanged.
2. `chat_message_list_visible` excludes hidden; `chat_message_list` includes it.
3. A hidden message lands in `_messages_for_llm` but not in `thread_render_items` / the snippet.

Minor/optional: `role` is unvalidated (a typo flows to litellm) — could tighten to `Literal["user","assistant","system"]`.

mitch-mitchel (approved):
good with the comment for notes

nadahlberg:
@mitch-mitchel good call on tests. Added.

I think I'll skip the validation for role. This gets validated downstream on ChatMessage.save because ChatMessage.data is a json schema field that validates message data (not just role, but also what other fields can accompany a given role). Since message structure might change as over time, I'd rather enforce this at a single point.

Issue #518, opened 2026-06-11
Topic Flow: support links (and basic rich text) in info section bodies


mitch-mitchel:
## Problem

Topic Flow `info` section bodies render as plain text (`{{ body|linebreaks }}`) — no links, no markup. Surfaced while building the real ND corpus: official-resource URLs (ndcourts.gov, forms PDFs), and cross-track links ("use the waiver-track page instead"), can only sit as unclickable text. The only place a link renders today is a vCard contact's `url`, which is a workaround.

## What I'd like

Allow safe links (and possibly minimal rich text — bold/lists) in `info` bodies. Options to weigh: a constrained markdown subset rendered + sanitized server-side, vs. a structured "links" field on the section. Either way the rendered output must be sanitized/escaped (corpus is authored, but courts will author too — treat as untrusted).

## Considerations

- Sanitization / XSS — never render raw author HTML; whitelist.
- CSP (no inline handlers), WCAG (real `<a>` semantics, external-link affordance), no-JS (server-rendered).
- Relates to the structured official-resources idea (sibling issue) and #373 (Official Resources URLs on the AI prompt side).

## Definition of done

- An `info` body can include at least clickable links, sanitized and CSP/WCAG-clean.

PR #584, opened 2026-06-30
feat(topic-flow): first-class official resource links

### LINKED ISSUE #519: Topic Flow: first-class official/jurisdictional resource links for a flow
## Problem

Key jurisdictional URLs for a flow — the court self-help center, official forms/PDFs, relevant statutes — have no first-class home in a Topic Flow corpus. Today they can only be embedded as plain text in `info` bodies (which don't render links). This is the engine-side counterpart to #373, which adds Official Resources URLs to the AI chat prompts.

## What I'd like

A corpus-level `resources` (or "official links") list — modeled like `contacts`/`deadlines` (id + label + url + optional note), referenced by an output/section and rendered as a labeled link list. Courts supply them as data (open-contracts), so the engine renders trustworthy official links without authors hand-coding HTML.

## Considerations

- Same validated-contract shape as contacts/deadlines (loader cross-checks references).
- Link rendering: real `<a>`, external-link affordance, WCAG; CSP-clean; works no-JS.
- Overlaps the "links in info bodies" sibling issue — these are two design directions (structured data vs free-text links); grooming can pick one or keep both.
- Relates #373.

## Definition of done

- A corpus can declare official resource links as structured data, rendered as clickable links in the flow.

mitch-mitchel:
Adds a corpus-level `resources` list (id + label + required url + optional note) and a `resources` output section that references them by id — mirroring the contacts/deadlines validated-contract pattern. The renderer resolves refs in declared order into a labeled link list (real `<a>`, external-link affordance, server-rendered, JS-off); the loader cross-checks every reference resolves and flags duplicate ids.

Both ND name-change tracks migrate to it: the official self-help and name-change guidance pages move from unclickable plain text in an info body into structured, clickable links — retiring the dead-text-URL workaround.

Resolve is inline in the renderer (no download artifact, unlike vcf/ics), and `url` is required on `Resource` (a resource *is* a link). #518 (inline links within info-body prose) stays open as the complementary follow-up.

Closes #519

## Test plan

- schema: `url` required, label/url/note carried, unknown-key rejected, empty `resource_ids` rejected
- loader: unknown-ref fails loud, duplicate-id detected
- renderer: resolves referenced resource, declared-order preserved, distinct template
- view: resources section renders a clickable `<a href=url>`
- `make lint` + `make test` green locally

### DIFF
--- a/litigant_portal/app/templates/cotton/molecules/flow_section_resources.html
+++ b/litigant_portal/app/templates/cotton/molecules/flow_section_resources.html
@@ -1,0 +1,17 @@
+{% load i18n %}
+
+{# resources output body: official/jurisdictional links for the flow (court #}
+{# self-help center, statutes, forms index), supplied by the corpus as data #}
+{# (#519). Server-rendered and JS-off — real <a> links, external-link icon + #}
+{# rel="noopener noreferrer" via the link atom. Replaces smuggling official #}
+{# URLs into contact cards or unclickable info-body text. #}
+<ul class="space-y-3">
+  {% for resource in ctx.resources %}
+  <li>
+    <c-atoms.link href="{{ resource.url }}" target="_blank" external_icon="true">{{ resource.label }}</c-atoms.link>
+    {% if resource.note %}
+    <p class="text-sm text-greyscale-500">{{ resource.note }}</p>
+    {% endif %}
+  </li>
+  {% endfor %}
+</ul>
--- a/litigant_portal/app/tests/test_topic_flow_loader.py
+++ b/litigant_portal/app/tests/test_topic_flow_loader.py
@@ -10,8 +10,10 @@ from litigant_portal.app.topic_flow.loader import (
     CorpusLoader,
     CorpusValidationError,
 )
+from litigant_portal.app.topic_flow.schema import ResourcesOutput
 
-FIXTURE = Path(__file__).resolve().parents[2] / "content" / "_test_fixture.yml"
+CONTENT = Path(__file__).resolve().parents[2] / "content"
+FIXTURE = CONTENT / "_test_fixture.yml"
 
 # A minimal schema-valid corpus: one fact_gather question, no deadlines or
 # outputs. Tests deep-copy and mutate this to introduce specific problems.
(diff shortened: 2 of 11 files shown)


mitch-mitchel:
Self-review — a few things I'd weigh before merge:

1. **Duplicate destination in the ND corpora.** The `self_help` contact and the new `nd_self_help_center` resource both point at `ndcourts.gov/legal-self-help`, so the same URL now renders twice on the page — once under "Save these contacts" (with phone/email), once under "Official resources". The name-change *guidance* page is the unique value the resources section adds; the self-help-center link is arguably redundant with the contact. Lean toward dropping `nd_self_help_center` and letting resources carry only the guidance page (and, later, the statute), unless we want the duplication deliberately.

2. **No test exercises the real ND corpora.** Every loader/registry test runs against `tmp_path` fixtures, so the new `resources` sections in `adult-name-change-{standard,waiver}.yml` are validated only by `make`/runtime — a typo'd `resource_id` wouldn't fail a unit test. Worth a small test that loads both real tracks and asserts the resources section resolves (would also guard the contacts/deadlines/packet sections already in there).

3. **Minor: repeated dict lookup in `_render_resources`.** It does `by_id[ref]` four times per ref; `contacts.py` (the resolver this mirrors) binds `contact = by_id[id]` once. Bind `resource = by_id[ref]` for parity/readability.

4. **Cross-cutting a11y (not a blocker for this PR).** Resource links open a new tab with only an aria-hidden external icon — no sr-only "opens in new tab" cue. Same precedent as the packet form links and vcf contact URLs, so it's a shared follow-up rather than something to fix only here.

Nothing here is load-bearing for correctness — the engine path is solid and covered. (1) and (2) are the two I'd actually action.

mitch-mitchel:
Addressed in `73c39c5`:

1. **Duplicate destination** — dropped `nd_self_help_center` from both tracks; resources now carry only the unique name-change guidance page (the self-help center stays as the `self_help` contact, where the phone/email belong).
2. **Real-corpus coverage** — added a parametrized test that loads the actual `adult-name-change-{standard,waiver}.yml` and asserts the resources output is wired, so a dangling `resource_id` fails CI, not just at runtime.
3. **Repeated lookup** — `_render_resources` now binds `resource = by_id[ref]` once, matching `contacts.py`.

Deferring **(4)** the new-tab a11y cue — it's shared with the existing packet/vcf links (aria-hidden external icon, no sr-only "opens in new tab"), so it belongs in a cross-cutting follow-up rather than this PR. Can file it if we want it tracked.

Issue #578, opened 2026-06-29
Track llm usage


nadahlberg:
Add fields to ChatMessage for token estimation and cost. Add helper selectors to aggregate costs across a thread.

PR #586, opened 2026-06-30
Track llm usage

nadahlberg:
This PR does a few things:

- First we tweak how the model is passed in the agent engine. This should be set at the endpoint-level (currently hardcoded, will eventually be inherited from database admin config models). As a guard, if the developer tries to set `model` in the agent's completion args we will raise an error. We update the docs and the demo agent to reflect this.
- Second, a chat_message_create helper is added. This wraps ChatMessage.objects.create, and if no num_tokens is provided it will be calculated. The reason for this is: (i) For user messages we have to calculate num tokens manually, but for (ii) assistant messages and tool calls we want to rely on the litellm response `usage` object, since this factors in tool call args and thinking tokens and is the most accurate cost tracker.
- The costs of llm messages are automatically passed, again since the litellm response object will include the most accurate cost tracking. User messages don't have an associated cost, so these default to zero.

Seeing the usage on the frontend:
- We have a chatUsage component that only gets rendered on the frontend for developers (superusers).
- We also add a corresponding endpoint that this component pulls data from

So if you are logged in with a superuser account, you'll see something like this:
Screenshot 2026-06-30 at 4 17 58 PM

We also update docs, old tests with the new model pattern, and add new tests.

### DIFF
--- a/litigant_portal/app/services/chat_v2.py
+++ b/litigant_portal/app/services/chat_v2.py
@@ -243,10 +243,17 @@ def chat_stream(
                 for chunk in litellm.completion(**call_args):
                     usage = getattr(chunk, "usage", None)
                     if usage:
-                        cost = litellm.completion_cost(
-                            completion_response=chunk, model=model
-                        )
-                        completion_tokens = usage.completion_tokens
+                        try:
+                            cost = litellm.completion_cost(
+                                completion_response=chunk, model=model
+                            )
+                            completion_tokens = usage.completion_tokens
+                        except Exception:
+                            logger.exception(
+                                "Chat cost/usage extraction failed "
+                                "(model=%s)",
+                                model,
+                            )
                     if not chunk.choices:
                         continue
                     delta = chunk.choices[0].delta
--- a/litigant_portal/app/services/chat_v2.py
+++ b/litigant_portal/app/services/chat_v2.py
@@ -250,8 +250,7 @@ def chat_stream(
                             completion_tokens = usage.completion_tokens
                         except Exception:
                             logger.exception(
-                                "Chat cost/usage extraction failed "
-                                "(model=%s)",
+                                "Chat cost/usage extraction failed (model=%s)",
                                 model,
                             )
                     if not chunk.choices:
--- a/litigant_portal/app/services/chat_v2.py
+++ b/litigant_portal/app/services/chat_v2.py
@@ -20,18 +20,42 @@ logger = logging.getLogger(__name__)
 MAX_STEPS = 30
 
 
+def chat_message_create(
+    *,
+    thread_id,
+    data: dict,
+    model: str,
+    num_tokens: int | None = None,
+    hidden: bool = False,
+    cost: float = 0.0,
+) -> ChatMessage:
+    """Add a message to a thread."""
+    if num_tokens is None:
+        num_tokens = litellm.token_counter(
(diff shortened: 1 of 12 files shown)


nadahlberg:
Ultimately this makes it so that cost and token count tracking are automatic whether chat turns take place via the `chat_stream` method, or when messages are manually injected via `chat_message_create` or `chat_message_inject_hidden`

nadahlberg:
Oh and fixes #578

litigant_portal/app/services/chat_v2.py
```diff
@@ -194,16 +227,28 @@ def event_stream() -> Iterator[str]:
...
 
                 content_parts: list[str] = []
                 tool_calls: list[dict[str, Any]] = []
+                cost = 0.0
+                completion_tokens = None
 
                 for chunk in litellm.completion(**call_args):
+                    usage = getattr(chunk, "usage", None)
+                    if usage:
+                        cost = litellm.completion_cost(
```
mitch-mitchel: `completion_cost` runs inside the chunk loop under the broad `except Exception`. If it raises — model not in litellm's pricing map (`gpt-5-mini` may be unpriced in the pinned version), or it can't cost a bare stream chunk — the turn aborts **after** content already streamed, and the assistant `chat_message_create` below (line 290) never runs, so the reply is lost from history on reload. Suggest wrapping just the cost/usage extraction in its own try/except, defaulting `cost` to 0.0. Dev-only telemetry shouldn't be able to break the stream.

litigant_portal/app/services/chat_v2.py
```diff
@@ -194,16 +227,28 @@ def event_stream() -> Iterator[str]:
...
+                            completion_response=chunk, model=model
+                        )
+                        completion_tokens = usage.completion_tokens
```
mitch-mitchel: `num_tokens` stores `completion_tokens` (output only). Prompt tokens — the full resent history + system prompt + tool schemas, every turn — are never recorded, so `chat_thread_usage` total_tokens understates real consumption, while `cost` (from `completion_cost`, prompt+completion) is accurate. The superuser readout then pairs a correct $ with a token count that doesn't back into it. If that's intentional (per-message stored tokens, not billed), a clarifying label would help.

litigant_portal/app/static/js/chat_v2.js
```diff
@@ -623,4 +652,43 @@ document.addEventListener('alpine:init', () => {
...
+  // superusers (so non-superusers never instantiate it or hit the endpoint).
+  // Assumes it lives inside a chatApp component — it reads the thread id from
+  // the parent and refreshes when the thread changes or a turn finishes.
+  Alpine.data('chatUsage', () => ({
+    totalTokens: '0',
+    totalCost: '$0.00',
+    app: null,
+
+    init() {
+      this.app = this.$root.closest('[x-data="chatApp"]')._x_dataStack[0]
```
mitch-mitchel: Minor: reaches into Alpine's private `_x_dataStack`. `Alpine.$data(el)` is the public equivalent and survives Alpine-internal refactors on upgrade — swap if it's exposed in the CSP build.

mitch-mitchel (commented):
Tidy PR — the model guard, the `num_tokens is None` check, hidden-message inclusion, and superuser gating on both endpoint and template are all solid. Two things worth a look plus one minor note (inline).

PR #622, opened 2026-07-07
feat(middleware): add middleware for temp site password

nadahlberg:
This PR adds a temporary middleware to block the whole site unless you have a password. Affects views only, not api endpoints so that health checks will still work.

I also snuck in a few small fixes so the staging deploy goes more smoothly:
- Fixed the default chat model with a valid bedrock slug for haiku
- Added optional env vars for testing bedrock models locally with your aws session credentials
- Deleted the S3 env vars from the docker compose since dev uses filesystem storage

### DIFF
--- a/litigant_portal/app/templates/site_password.html
+++ b/litigant_portal/app/templates/site_password.html
@@ -1,0 +1,29 @@
+{% load static %}
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="robots" content="noindex">
+    <title>Password required</title>
+    <link rel="stylesheet" href="{% static 'css/main.built.css' %}">
+</head>
+<body class="min-h-dvh flex items-center justify-center bg-greyscale-50">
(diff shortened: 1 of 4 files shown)


nadahlberg:
@mitch-mitchel here's the password guard. If this looks good, shall we merge this and then merge main into staging (and cross our fingers and see what happens)

litigant_portal/app/templates/site_password.html
```diff
@@ -0,0 +1,29 @@
...
+    <form method="post" class="w-full max-w-sm p-6 bg-white border border-greyscale-200 rounded-lg shadow-sm">
+        <h1 class="text-lg font-bold mb-4">This site is under development</h1>
+        <p class="text-sm text-greyscale-600 mb-4">
+            This site is a work in progress and not yet open to the public.
+            If you're part of the project, enter the password to continue.
+        </p>
+        {% if error %}
+            <p class="text-sm text-red-700 mb-4">Incorrect password.</p>
+        {% endif %}
+        <input type="password" name="site_password" autofocus
```
mitch-mitchel, line 21: nit, but can you just add User/Pass display text

mitch-mitchel (approved):

PR #629, opened 2026-07-07
feat(middleware): return 200s for site password page

nadahlberg:
The password guard is returning 401s. Changing this to 200s since the guard is just to block user's from content, not a real auth error. And the 401s are confusing our EKS probes


nadahlberg:
Actually I'm sneaking in another commit to temporarily disable the custom domain for s3. Will open an issue to revert once the storage DNS is setup

Issue #633, opened 2026-07-07
feat: persistent Topic Flow links in chat (deterministic AI to flow handoff)


mitch-mitchel:
### Problem

The chat → Topic Flow handoff is prompt-driven — the LLM has to emit a markdown link — so it fires inconsistently (#602) and hardcodes a QA-absolute URL. A deterministic navigation affordance should not depend on a non-deterministic LLM.

### Fix

Render the Topic Flow track links for the topic as **persistent, server-rendered UI** in the case/briefcase panel — present from the moment you enter the topic, derived from the corpus registry, not from LLM output. The LLM prompt is reworded to *reference* the on-page links (and can still recommend the right track), but never has to emit a URL.

**Court resolution (omni-court):** the topic resolves the court — each topic carries the real corpus of its court (name change → ND, eviction → Franklin County). Topic has a corpus → show that court's track links (e.g. ND name change → **Standard** / **Waiver**); no corpus → generic, no links. No session court state needed (#632 dropped; in prod the wiki corpus feed supplies court context).

**Implementation** (design carried from planning):
- New `registry.tracks_for(topic)` in `app/topic_flow/registry.py` resolves the owning court + available `role` tracks for a topic from `content/*.yml` (bridges the `adult_name_change` chat slug ↔ `adult-name-change` corpus slug in one documented place).
- `chat_page` (`app/views/pages.py`) passes `flow_tracks` to the template.
- Render via `<c-molecules.sidebar-section>` + `<c-atoms.link>` inside the case-sidebar, using `{% url 'pages:topic_flow' court topic role %}` (relative — also fixes #600).
- Rewrite `prompts/courts/north-dakota/prompt.md` + `prompts/topics/adult_name_change/prompt.md` to reference the on-page links instead of emitting hardcoded QA URLs.
- Build in v1 chat (`pages/chat.html`); the registry helper + prompt change carry over to v2.

### Supersedes

The prompt-only approach in #602. Closes #602 when this lands.

Critical-for-day-1: gives Jessica demos + court partners reliable access to the Topic Flows that already work.

Issue #632, opened 2026-07-07
feat: court context — demo hot-switcher (Jessica/partners) + prod per-court config


mitch-mitchel:
Establish a **court context** for a session so Topic Flows, chat→flow handoff links, and (later) court branding resolve against the right court. Two modes:

- **Demo — a court hot-switcher.** A UI affordance so Jessica and court partners can flip the active court on the fly during demos and testing, without a separate deploy per court.
- **Prod — a per-court config object.** In a real court deployment the court is fixed by a config object (per partner), not a user-facing switcher.

### Why

Court-partner testing and Jessica demos need to operate *inside a court* from day one. This is the prerequisite for the deterministic AI → Topic Flow handoff links resolving the correct court's flows, and it underpins court-partner branding (#365, #306) and court-configured AI context (#179).

Today the main entry (`/chat/?topic=…` from a topic card) carries no court, so nothing court-specific has a court to resolve against.

### Scope / to decide

- Where the demo switcher lives (header affordance? demo-only visibility?) and how it sets the active court (session-scoped).
- The prod config-object shape and how court resolves with no switcher present.
- Relationship to URL scoping (#365 `/{court}/` demo-site URLs) — session vs URL as the source of truth for the active court.

Critical-for-day-1: Jessica demos + court-partner testing/feedback depend on it.

mitch-mitchel:
Dropped — omni-court: the topic resolves the court (each topic carries its court's real corpus), and the wiki corpus feed supplies court context in prod. No switcher, no per-court config object. #633 re-scoped accordingly.

PR #635, opened 2026-07-08
feat(chat): demo court context switcher (dev/QA only)

mitch-mitchel:
Adds a session-backed active court with a hot-switcher in the dev-tools menu, so demos and court-partner testing can operate inside a court from the start. Production resolves court from a per-court config object instead (future), so the switcher is hidden and inert outside dev/QA. This is the substrate the persistent chat→flow links (#633) build on via `get_active_court`.

- `court_context` helper: one validated path for reading/writing the active court (rejects unknown/stale slugs; clears on the generic option)
- context processor exposes switcher state + active court app-wide
- `set-court` POST view: dev/QA-gated, safe same-host redirect back
- switcher `<select>` in the existing dev-tools menu (already gated `deployment_env != "prod"`)
- `chat_page` resolves court from the session; an explicit valid `?court=` (deep-link) overrides and persists it

Part of #632

## Test plan
- `make test` — new `test_court_context.py` (helper + context-processor DB-free `SimpleTestCase`; `set-court` view + `chat_page` behavioral) is green; retargeted two pre-existing `test_portal.py` court-branding tests to the `court-branding` eyebrow, which the switcher's dropdown options no longer collide with.
- Manual (dev/QA): header dev-tools menu → "Active court" → Switch → the choice persists across navigation and the sub-header branding reflects it; confirm the switcher is absent under `DEPLOYMENT_ENV=prod`.

### DIFF
--- a/litigant_portal/app/context_processors.py
+++ b/litigant_portal/app/context_processors.py
@@ -1,12 +1,31 @@
 from django.conf import settings
 from django.contrib.messages import get_messages
 
+from litigant_portal.app import court_context as court_ctx
+from litigant_portal.prompts import get_court_name
+
 
 def app_meta(request):
     """App-level metadata available in every template."""
     return {
         "deployment_env": settings.DEPLOYMENT_ENV,
         "app_build_time": settings.APP_BUILD_TIME,
+    }
+
+
+def court_context(request):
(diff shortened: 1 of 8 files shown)


mitch-mitchel:
holding off - omni court with separate topics good for now

PR #706, opened 2026-07-29
chore(deps): drop orphaned pdfplumber dependency

### LINKED ISSUE #690: Drop orphaned pdfplumber dependency
`pdfplumber`'s sole importer was `services/pdf_service.py`, deleted in #682. The live attachment pipeline (`services/attachments.py`) uses `pypdf`/`mammoth`/`openpyxl`/`tiktoken`. pdfplumber and its transitive deps (pdfminer.six et al.) are dead weight built into every image.

**Fix:** remove `"pdfplumber>=0.11"` from `pyproject.toml` and run `uv lock` (the pre-commit uv-lock-check hook requires the lock be regenerated).

mitch-mitchel:
pdfplumber was chat v1's PDF text-extraction dep; the v2 attachment pipeline reads uploads with pypdf, and the v1 cleanup (#682) removed its last importer. Dropping it also removes transitive Pillow 12.2.0, clearing all 13 open Dependabot alerts at the root instead of committing to version bumps for an unused library.

Closes #690

Test plan: `make pre-commit` (lint + full suite) passes locally; grep confirms zero `pdfplumber`/`PIL` imports in `litigant_portal/`; pillow and pdfplumber both absent from `uv.lock`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,6 @@ dependencies = [
     # Rate limiting
     "django-ratelimit>=4.1",
     # PDF text extraction
-    "pdfplumber>=0.11",
     "pypdf>=5.0",
     "mammoth>=1.8",
     "openpyxl>=3.1",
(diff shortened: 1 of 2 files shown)


nicolearagao (approved):
LGTM

Issue #684, opened 2026-07-21
Delete orphaned services/search_service.py (unimportable since #682)


nadahlberg:
`litigant_portal/app/services/search_service.py` imports the `Document` model, which migration 0010 deleted — the module no longer imports at all. Its only caller (the v1 fallback-search endpoint) and its tests were archived with chat v1, so nothing can catch it: zero live importers remain.

**Fix:** delete the file (or move it to `archive/v1_code/` alongside its already-archived tests).

Issue #742, opened 2026-08-03
User facing topic flow page


nadahlberg:
- Rewrite the litigant-facing flow page against the models; drop the court/topic/role URL for topic/flow.
- Answer save/load per UserIdentity, packet/ICS/VCF generation off the model data.
- Cleanup old topic flow module and template

Issue #731, opened 2026-08-03
Enforce Site as a singleton model


nadahlberg:
We no longer need multiple site configs. Now Site will be a singleton model -- it always has exactly one row.
- Pin Site.id to a fixed UUID with a CheckConstraint; drop the active field and its unique constraint.
- post_migrate guarantees the row exists.
- site_get() cached under one key; writes wear @busts_site_cache (commit-time invalidation).
- Remove the developer site-switcher from the settings tab.

PR #743, opened 2026-08-03
Regroup and rename the service/selector/model layer

### LINKED ISSUE #729: Regroup and rename the service/selector/model layer
Reorganize services, selectors, and models to group by domain and add more consistent method naming.

General idea: models and related utility functions tend to get used across pages more often than endpoints. So group models and their utilities by domain. Here are the domains that stand out:

- `site`: stuff related to the global site settings and court config. This is the stuff currently in admin, but site settings will apply across the app, so better to give it its own namespace at the service layer.
- `chat_engine`:  chat engine stuff. Probably good where it is.
- `topic_flow`: stuff related to the topic flow engine. Another reason to move admin to site, because the topic flow utilities span both the admin site and user-facing topic flow page. Let it all live here.
- `user`: move identity related stuff to a user module. This will be a good place to support both identity and things like update user profile helpers.
- `libary`: this won't exist yet, but will be the place for logic related to importing configs from the library
- `upload`: currently in the assistant module. The attachment upload has a lot of helper logic, let it live in its own module.

Apply consistent naming to services and selector methods. A little liberty can be afforded when a utility genuinely implicates more than one model equally, but generally these should be named like `{model_name}_{action}`. 

Group views and endpoints by surface (the specific page they apply to):
- `pages`: This is used for views that actually render a page. Keep these very thin, let endpoints do the work to keep pages reactive.
- Endpoint modules: create one endpoint module per page. Define the endpoints used by that page here.

In urls, maintain separate lists for pages and the various endpoint groups.

nadahlberg:
Fixes #729, #684

Groups models, selectors, and services by **domain** rather than by the surface that happens to consume them, and applies consistent `{model_name}_{action}` naming throughout. Views and endpoints are grouped by **surface**.

No schema changes, no route changes — every URL pattern name is byte-identical to `main`. It lands first because it touches nearly every module in the app, and doing it ahead of the feature work keeps the branches that follow from fighting each other over import lines.

### Why

Models and their utilities get reused across pages far more often than endpoints do, so the layer below the view should be organized around the data, not around who calls it. The `admin` module was the clearest symptom: it held site settings, topic CRUD, and user permission checks, none of which are actually about the admin page. Site settings apply app-wide; topic utilities span both the admin site and the user-facing topic flow page.

### Domains

| Module | Holds |
| --- | --- |
| `site` | Global site settings and court config (was `admin`) |
| `topic_flow` | Topic and topic flow data, spanning admin and public surfaces |
| `user` | Identity plus the permission predicates and toggles |
| `upload` | Attachment upload and its substantial helper logic (was `assistant`) |
| `chat_engine` | Unchanged — already correctly scoped |

### Naming

Services and selectors are now `{model_name}_{action}`:

- `identity_ensure` → `user_identity_ensure`, and likewise across the merge helpers
- `attachments_for_llm` → `user_upload_llm_parts`
- `reader_limit_error` → `user_upload_reader_limit_error`
- `content_part` → `user_upload_content_part`
- `attachment_render_list` → `user_upload_render_list`
- `thread_render_items` → `chat_thread_render_items`
- `site_get_active_topics` → `topic_list_active`

The upload module's internal helpers (`_file_kind`, `_extract_text`, `_pdf_page_count`, `_hydrate_attachment`, and friends) are now underscore-private, so the module's public surface is just the `user_upload_*` functions.

### Reads are selectors, writes are services

Permission predicates included. `user_can_access_admin`, `user_is_developer`, and `user_can_manage_site` are reads, so they moved to `selectors/user.py` alongside `user_list`; `user_developer_toggle` writes, so it stays in `services/user.py`. Before this, `context_processors.py` imported a *service* to answer a yes/no question.

`site_membership_toggle` moved to `services/site.py` to sit with the model it writes.

### Cache keys follow the data

Splitting `admin` left the site and topic_flow modules pointing at each other: `selectors/site` imported `topic_list` from `selectors/topic_flow`, while `services/topic_flow` reached back into `selectors/site` for the cache key it needed to bust on every write.

`site_get_active_topics` returned topic rows, so it's now `topic_list_active` in `selectors/topic_flow`, which also owns `ACTIVE_SITE_TOPICS_CACHE_KEY`. The dependency runs one way again: `services/site.site_activate` imports the key to invalidate it, and nothing points back. The cache key *string* is unchanged, so no deployment cache flush is needed.

The general rule, now in CLAUDE.md: a cache key lives in the selector module that reads through it, and whoever invalidates it imports it from there.

### Views and URLs

`pages` holds views that render a page. Each surface gets its own endpoint module alongside it — `admin`, `assistant`, `topic_flow` — and `urls.py` keeps a separate pattern list per group.

### Model surface

`models/__init__.py` re-exports explicitly instead of `from .x import *`. Eight names, one list, checked by ruff instead of by `# noqa: F403`. It also settles an ambiguity this PR would otherwise have introduced: `models/topic_flow.py` imports `Site` for its FK, and under the old star-imports that made `Site` arrive at the package from two different modules.

### Not a pure move

Two changes go beyond relocating code:

- **`views/chat_engine.stream`** — the attachment ownership check now goes through the `user_upload_list` selector instead of querying `UserUpload` directly. Same count, correct layer.
- **Migration `0007`** — hand-edited, because `upload_to=` serializes the callable's module path and the callable moved with the model. Without the edit the migration won't import and `makemigrations` sees drift. `makemigrations --check` is clean.

### Renamed

Filenames now track the modules they cover:

- `tests/test_attachments.py` → `tests/test_upload.py`
- `tests/test_identity_services.py` → `tests/test_user_services.py`
- `docs/ai-tooling/ATTACHMENT_SYSTEM.md` → `docs/ai-tooling/UPLOAD_SYSTEM.md`

### Removed

- `services/search_service.py` — dead since the search work moved on; nothing outside `archive/` imported it
- `llm_completion` in `agents/message_schema.py` — a litellm wrapper with no remaining callers

### Two things sharing a name

`app/topic_flow/` is the corpus engine (schema, registry, renderer, downloads) that reads YAML out of `litigant_portal/content/`. The `topic_flow` modules under `models/`, `selectors/`, `services/`, and `views/` are the data layer for the `Topic` rows that corpus attaches to. The former is the old topic flow engine that will ultimately be replaced by the latter in future PRs.

### Disclaimer

Three things that break the patterns but which will go away imminently:

- `topic_flow_download` moved into `views/topic_flow.py` as an endpoint but keeps its `pages:`-namespaced URL. Renaming it would change the route, which this PR avoids. Resolved by #742, which deletes the view outright.
- `services/site.py` imports the topic-flow cache key so `site_activate` can invalidate it. Resolved by #731 — `site_activate` goes away with `Site.active`.
- `site_get_active_data` and `site_get_active` are redundant readers of the same row. Resolved by #731, which collapses them into one cached `site_get()`.


mitch-mitchel:
@nadahlberg  few nits/requests
- also closes #684, please add
- for next branch, can you use our convention ex: for this PR would have been 729-{slug}
- can you add a migration comment for the hand edit for our future selves - asked claude recmomend something like this
```
  # Hand-edited: models/assistant.py became models/upload.py (#729). FileField's
  # upload_to serializes the callable's import path, so this reference had to
  # follow the move or the migration won't import and makemigrations sees drift.
  # Safe to edit in place — upload_to is Python-side only, never stored in the
  # DB, so no schema change and no data migration.
  ```
  - current pattern violations that go away - can you tag the issues that fix those?

File: litigant_portal/app/urls.py [hunk 1]
```diff
@@ -15,6 +15,9 @@
     health,
     pages,
 )
```
mitch-mitchel, line 17: this should just be one block, right?

mitch-mitchel (dismissed):
small cleanup, 

sorry, that text selection did not land - this same imprort on line 13 is in 2 blocks, think it can just be one
`from litigant_portal.app.views import`

nadahlberg:
@mitch-mitchel added the migration comment and issue refs in the PR message.

litigant_portal/app/urls.py
nadahlberg: Not with the current ruff settings (pre-commit will just undo). We can add `combine-as-imports = false` in a different PR if you want these consolidated.

mitch-mitchel: interesting, got it

mitch-mitchel (approved):

nicolearagao (approved):
Wasn't familiar with some of the patterns existing at Django-Styleguide repo. 

LGTM

Issue #621, opened 2026-07-07
We haven't established why we need user's information before forms


JessicaFrank:
What's the plan for the "Your details" fields? What are they used for?

Are they reused in the forms? Do they get stored in the user's answer file? 

In anonymous workflow here, the user isn't told why we're asking for this information. 

Same goes for Summary of what you entered section. What's the point? Are we building out a the briefcase?

If we don't actually use this information anywhere, we shouldn't ask for it. The form's guided interview will capture it. If it's useful for some other reason, we need to buffer the fields with an explanation of why we're gathering this. 

If we gather it, we should also be able to reload it into the form's guided interview so the same information isn't asked of the user multiple times. 

Optimize for the fewest times of re-entering data as possible.

mitch-mitchel:
The intent was to do a hand off to the docassmble flow, but this the wrong space. It should either pull from the main portal login user info, or be pure anonymous at the Topic Flow level. removing.

Issue #515, opened 2026-06-11
Topic Flow: user-facing answer clearing — per-answer clear + start over


mitch-mitchel:
## Problem

In a Topic Flow `fact_gather` section, once a litigant saves an answer there's no way to *clear* it — only overwrite it with something else. A flow is usually only ~3 fields, and people change their minds or enter something by mistake, so they should be able to remove a single answer or wipe the whole flow and start fresh. Today neither is possible from the UI.

(This is the user-facing need behind the audit's "Reset Demo doesn't clear flow answers" note — the fix is real user controls, not the dev-only demo reset. The dev nuke is intentionally out of scope here.)

## What I'd like

Two controls, both server-driven (no JS required):

1. **Clear a single answer** — a control per answered field to remove just that answer (e.g. a "Clear" affordance on/next to each field in the `fact_gather` form). Backend: a per-question delete on `AnswerStore`.
2. **Start over** — clear all answers for this flow and return to the empty form. Backend already exists: `AnswerStore.clear()` drops just this `(court, topic, role)` flow, leaving any other flows untouched.

## Design considerations

- **No-JS / progressive enhancement (org floor):** both actions are form POSTs, not JS-only. Reuse the PRG + return-to-section pattern (#510) so clearing lands the user back at the `fact_gather` section, not the page top.
- **Destructive-action guardrail:** "Start over" wipes all answers — confirm before it fires (ties to #468). Per-answer clear is low-stakes; likely no confirm needed.
- **WCAG:** clear/start-over controls clearly labeled (not icon-only without a name); confirm dialog reachable and announced.
- **CSP:** no inline handlers; Alpine CSP build only if any enhancement is added.
- **Guest session-backed**, same as the rest of the flow — no auth, no DB.

## Definition of done

- A litigant can clear an individual saved answer from the flow page; the field returns to empty.
- A litigant can "Start over" to clear all answers for the current flow (with a confirm), other flows untouched.
- Both work with JavaScript disabled and land the user back at the relevant section.
- Meets CSP + WCAG.

PR #747, opened 2026-08-04
fix(topic-flow): don't leak cached answers

### LINKED ISSUE #638: Previous answers are cached for user input fields
I think we've talked about this before and it was tied to it being a QA site with static prompts/fields cached, but we need to make sure it doesn't do this in production. 

Previous answers to fields are being cached (by the browser?) and then redisplayed as options in different sessions. While it could be a benefit in some cases, the risk of exposing prior users' answers in public terminal settings is too high. We have to expect that many users are going to be on public computers and will not want previous answers shown. 

So ensuring field data doesn't persist/cache is important. 

Image

I was an anonymous user for this screenshot (in Chrome) who had answered questions in this page during a demo yesterday. I didn't clear my browser cache or close out the tab before testing this.

nicolearagao:
Browser autofill and the flow's own session prefill both echoed a prior user's answers back into the form on a shared/public terminal (#638). The #621 prune already removed the riskier name/county questions, so publication_date was the one field still exposed.

- atoms/input.html was silently dropping the autocomplete prop that form_field.html already forwarded to it — wire it through, and set autocomplete="off" on fact_gather's text fields.
- renderer.py never echoes publication_date's stored value into the form; it's still saved to AnswerStore for deadline computation, just not re-displayed.

Per-field clearing is a bigger, separate change (#515) — this targets the one field actually affected today.

### DIFF
--- a/litigant_portal/app/templates/cotton/atoms/input.html
+++ b/litigant_portal/app/templates/cotton/atoms/input.html
@@ -8,6 +8,7 @@
        {% if id %}id="{{ id }}"{% endif %}
        {% if value %}value="{{ value }}"{% endif %}
        {% if placeholder %}placeholder="{{ placeholder }}"{% endif %}
+       {% if autocomplete %}autocomplete="{{ autocomplete }}"{% endif %}
        {% if disabled %}disabled{% endif %}
        {% if readonly %}readonly{% endif %}
        {% if required %}required{% endif %}
--- a/litigant_portal/app/templates/cotton/molecules/flow_section_fact_gather.html
+++ b/litigant_portal/app/templates/cotton/molecules/flow_section_fact_gather.html
@@ -29,6 +29,7 @@
     name="{{ q.id }}"
     id="{{ q.id }}"
     value="{{ q.value }}"
+    autocomplete="off"
     help_text="{{ q.help_text|default:'' }}"
     :required="q.required"
     :autofocus="q.autofocus"
--- a/litigant_portal/app/tests/test_topic_flow_renderer.py
+++ b/litigant_portal/app/tests/test_topic_flow_renderer.py
@@ -91,6 +91,23 @@ def test_fact_gather_unanswered_question_prefills_empty():
     assert q["value"] == ""
 
 
+def test_fact_gather_never_prefills_publication_date():
+    # #638: unlike other fields, publication_date must never echo a stored
+    # answer back into the form.
+    section = _fg(
+        [
+            Question(
+                id="publication_date", label="Publication date", type="date"
+            )
+        ]
+    )
+    rendered = render_section(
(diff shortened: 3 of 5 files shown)


litigant_portal/app/templates/cotton/atoms/input.html
```diff
@@ -8,6 +8,7 @@
        {% if id %}id="{{ id }}"{% endif %}
        {% if value %}value="{{ value }}"{% endif %}
        {% if placeholder %}placeholder="{{ placeholder }}"{% endif %}
+       {% if autocomplete %}autocomplete="{{ autocomplete }}"{% endif %}
```
nicolearagao, line 11: Hey @mitch-mitchel ,  wanted to check I scoped #638 right.

Following your comment (browser autofill + our own session prefill, both ours, only publication_date left after the #621 prune): I tried to actually reproduce the browser-autofill dropdown in Chrome/Brave and couldn't. Turns out <input type="date"> doesn't get that suggestion-dropdown UI at all in Chromium browsers, since the native date widget doesn't hook into the classic autofill-history dropdown. So autocomplete="off" doesn't have a visible effect on this specific field, even though it's still the correct thing to set.

The part that actually reproduces the original screenshot is our own session prefill: renderer.py was echoing answers.get(q.id, "") for publication_date on every GET, so on a shared/uncleared browser, the next person sees the previous person's date already filled in.

One thing I initially missed and that was highlighted by Claude's review: the same leak existed a second time, in a different section. The recap/summary block near the bottom of the page (_answered_in_corpus_order → flow_section_summary.html) reads from the exact same session-backed answers and echoes publication_date verbatim into its <dd>,  completely independent of the fact_gather form fix, since it's a separate render path. So fixing the input field alone left the date still visible a few screens down. I've applied the same guard there too, so both paths are covered now.

What I landed on:

- publication_date no longer prefills from AnswerStore, ever,  neither in the form field nor in the recap summary. It's still stored for the deadline calc, just never echoed back anywhere on the page.
- Added autocomplete="off" to fact_gather fields anyway, as a correct default even though it's not visibly testable for this field type.
- Left #515 (per-field clear / start over) alone,  that's an opt-in action the litigant has to click, so it doesn't cover the "shows up by default with no action" risk #638 is actually about.

Does that match what you had in mind, or did you want this broader (e.g. no fact_gather field ever prefills anywhere on the page, not just publication_date)?

mitch-mitchel: That's the right fix - this answerStore will likely go away or be superseded by the larger context we chatted about. The happy path is a logged in user's info is shared to any Topic Flows and into any docassemble interviews that might be attached to that Topic Flow. This is a bigger architecture/design fix though that is TBD on delivery date, so this fix will cover us until then.

mitch-mitchel (approved):
🚢

PR #748, opened 2026-08-04
Replace SiteMembership with Django permissions

### LINKED ISSUE #730: Replace SiteMembership with Django permissions
We used the site membership model initially because we were supporting multiple site "configurations" so you could switch between Court X mode vs Court Y mode. We are no longer supporting that and will be making the Site model a singleton model, so that there is a single set of config options per app-deploy. This unlocks our ability to simplify our permission system and just use django native permission and group models.
- Delete the SiteMembership model and both custom context processors.
- Add manage_site / manage_developers as Site.Meta.permissions.
- Add Admins (manage_site) and Developers (both) groups via a post_migrate receiver.
- Replace the context processors with admin_access_required / developer_required view decorators.
- Keep these distinct from Django-admin permissions.

nadahlberg:
Fixes #730

Admin access is now Django's own permission framework. `SiteMembership` is gone, along with the two custom context processors and the predicate helpers that fronted them. There is no bespoke permission layer left — `user.has_perm(...)` is the whole API.

### The two permissions

Declared on `Site.Meta.permissions`:

| Permission | Who it's for | Grants |
| --- | --- | --- |
| `app.manage_site` | Court partners and administrators running the site | Access to the admin panel and everything configured there |
| `app.manage_developers` | Free Law Project developers working on the portal | Ability to grant and revoke developer access |

### The two groups

| Group | Who belongs to it | Holds |
| --- | --- | --- |
| `Admins` | Court partners and site administrators | `manage_site` |
| `Developers` | FLP developers working on the portal | `manage_site`, `manage_developers` |

`Developers` is a superset. FLP developers need the admin panel to build and support it, so they carry `manage_site` in their own right rather than also being added to `Admins`. Consequence worth knowing: removing someone from `Admins` does not revoke their admin access if they are also a developer.

The group names and the group-to-permission map live in `permissions.py`, above the selector/service split so both layers and the signal receiver can import them without importing each other.

A `post_migrate` receiver provisions both groups, so a fresh database, a restored dump, and a long-running deployment all converge on the same definition. It is additive — permissions granted to a group by hand are left alone. It warns rather than failing silently if the permissions it hands out don't exist yet, since the symptom would otherwise be a 403 for every admin.

### How permissions get checked

- **Page views** — Django's `@permission_required("app.manage_site", raise_exception=True)`, with `@login_required` outermost so an anonymous visitor is redirected to log in rather than shown a bare 403.
- **JSON endpoints** — `manage_site_required` / `manage_developers_required`, now in `views/utils.py` so any surface can use them. These stay custom because the built-in raises `PermissionDenied`, which renders an HTML 403, and these endpoints must answer with `{"error": "Forbidden"}` for the frontend to parse.
- **Templates** — `{% if perms.app.manage_site %}`, supplied by Django's auth context processor. The `admin_access` / `developer_access` context processors and their `settings.py` entries are deleted.

### Docs

New [`docs/wiki/permissions.md`](docs/wiki/permissions.md) covers the permissions, the groups, all three ways to check one, granting and revoking.

`CLAUDE.md` picks up a short permissions rule pointing at it, plus a line establishing that any package may carry a `utils.py` for helpers shared across its own modules.

### Behavior change worth reviewing

`user_can_manage_site` checked membership of *that specific site*. `manage_site` is global, so while multi-site still exists, any admin can now edit any `Site` row, and `site_list` no longer filters per user.

This is inherent to the swap — group permissions have no per-site scope — and #731 removes multi-site immediately after. Flagging it rather than burying it. Same reason `manage_developers` still gates `site_activate_view` and the site switcher alongside developer promotion: that endpoint disappears with multi-site, so it isn't worth re-homing now.

### No data backfill

The migration deletes `SiteMembership` without mapping existing rows to group membership, deliberately: the model was never deployed, so there is nothing to carry over. Anyone who needs access before the Users tab is reachable can be added to a group from Django admin.

### Tests

New `test_permissions.py`, 22 tests across five classes: the groups the receiver provisions, the toggle services, superuser access, the page guard, and the API guards. Negative cases are the point — no permission is a 403 on every admin endpoint, and both self-revocation guards are covered (an admin cannot drop their own admin access; nobody can drop their own developer status; a developer *can* toggle their own admin flag, since they keep `manage_site` through `Developers`). Superusers are pinned separately because `bootstrap_superuser` creates the account that has to be able to grant everyone else access, and it never touches a group.


CLAUDE.md
```diff
@@ -164,7 +164,9 @@ The compact rules (board mechanics live at the org level; sizing history, anchor
 
 **Naming:** services and selectors are `{model_name}_{action}` — `site_get_active`, `topic_create`, `user_identity_merge`, `user_upload_llm_parts`. Allow some liberty when a utility genuinely implicates two models equally. Anything not part of a module's public surface takes a leading underscore.
 
-**Reads are selectors, writes are services** — including permission predicates. `user_can_access_admin` reads, so it's a selector; `user_developer_toggle` writes, so it's a service. A cache key lives in the selector module that reads through it, and whoever invalidates it imports it from there.
+**Reads are selectors, writes are services.** `user_list` reads, so it's a selector; `user_developer_toggle` writes, so it's a service. A cache key lives in the selector module that reads through it, and whoever invalidates it imports it from there.
+
```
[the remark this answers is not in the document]
nadahlberg: Done

docs/wiki/permissions.md
```diff
@@ -0,0 +1,212 @@
+# Permissions
```
nicolearagao, line 1: nit: The scope in docs(docs) is redundant with the type. docs(permissions) would be clearer.

2c0b0eb - docs(docs): add permissions docs
8cadc71 - docs(docs): mention permission module in permission docs

litigant_portal/app/signals.py
```diff
@@ -1,8 +1,55 @@
...
+        group, _ = group_model.objects.using(using).get_or_create(name=name)
+        permissions = list(
+            permission_model.objects.using(using).filter(
+                codename__in=codenames, content_type__app_label="app"
+            )
+        )
+        if len(permissions) != len(codenames):
+            logger.warning(
+                "Group %r is missing permissions: expected %s, found %s. "
+                "Nobody in it will have admin access until they exist.",
```
nicolearagao: "they" could refer to the groups or the permissions. Since it's the permissions that are missing, "until the permissions exist" is unambiguous.

litigant_portal/app/signals.py
```diff
@@ -1,8 +1,55 @@
...
+from django.db.models.signals import post_migrate
 from django.dispatch import receiver
 
+from .permissions import GROUP_PERMISSIONS
 from .services.user import user_identity_merge_anonymous
 
+logger = logging.getLogger(__name__)
+
+
+@receiver(post_migrate)
```
nicolearagao: Could this be connected in AppConfig.ready() with sender=self instead? That way it only fires for this app and the sender guard on line 29 wouldn't be needed.

litigant_portal/app/services/user.py
```diff
@@ -1,18 +1,32 @@
...
 
 from litigant_portal.app.models import UserIdentity
+from litigant_portal.app.permissions import ADMINS_GROUP, DEVELOPERS_GROUP
 
 logger = logging.getLogger(__name__)
 
 
+def _group_toggle(*, user: User, name: str) -> bool:
+    """Flip a user's membership in a group; returns the new state."""
+    group, _ = Group.objects.get_or_create(name=name)
```
nicolearagao: If either group was deleted (human error, bad script), get_or_create silently recreates it as an empty group,  no permissions. The toggle returns True, meaning "done," but the user gets no access. That's a silent failure in the permission system.
Group.objects.get(name=name) would raise Group.DoesNotExist and fail loudly, which is the right behavior when the infrastructure is misconfigured. The ensure_permission_groups signal guarantees groups exist after every migration, so get is the appropriate call here.

litigant_portal/app/migrations/0011_site_permissions_delete_sitemembership.py
```diff
@@ -0,0 +1,22 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
nicolearagao, line 4: No data migration here for existing is_staff users, after this runs they lose admin access. Is that intentional? If there are real users in the database they'd need to be manually added to the Developers group, or this migration should move them over.

litigant_portal/app/views/admin.py
```diff
@@ -268,24 +238,22 @@ def _user_payload(user: User, *, viewer: User) -> dict:
...
         "joined": user.date_joined.strftime("%Y-%m-%d"),
-        "is_admin": getattr(user, "is_site_member", False),
-        "is_staff": user.is_staff,
-        "can_toggle_admin": viewer.is_staff or not is_self,
+        "is_admin": getattr(user, "is_admin_member", False),
+        "is_developer": getattr(user, "is_developer_member", False),
+        "can_toggle_admin": (
+            viewer.has_perm("app.manage_developers") or not is_self
+        ),
         "can_toggle_developer": not is_self,
```
nicolearagao: Should this be viewer.has_perm("app.manage_developers") and not is_self? The developer toggle endpoint requires manage_developers, so an admin would see the button but get a 403 when clicking it. The can_toggle_admin field above already does this check correctly.

nicolearagao (changes requested):

docs/wiki/permissions.md
nadahlberg: Makes sense -- will do this in future

litigant_portal/app/signals.py
nadahlberg: Fixed!

litigant_portal/app/services/user.py
nadahlberg: Oof, yes definitely. Fixed

litigant_portal/app/views/admin.py
nadahlberg: Yeah, it should. The entire column of buttons for toggling the developer status is guarded at the template-level, so non-dev admins won't see those buttons. But we should still fix the payload in case we ever use those variables somewhere else in the template.

Fixed.

litigant_portal/app/migrations/0011_site_permissions_delete_sitemembership.py
nadahlberg: This was intentional, just because we hadn't made any admin/dev users on the real DB yet. But wyt? Is it worth adding a backfill migration? I suppose it would help if any of us have stuff in our local dbs.

litigant_portal/app/signals.py
nadahlberg: Nice! Done

litigant_portal/app/migrations/0011_site_permissions_delete_sitemembership.py
nicolearagao: What do you think @mitch-mitchel ? I would be cool not having a backfill migration for this specific case, but do we want to have one anyway to establish good patterns ?

mitch-mitchel: i'm ok with not having it this time. once we have our baseline, we can lock this down @nicolearagao @nadahlberg

nicolearagao: The man has spoken =) Resolving this!

nicolearagao (approved):
Thanks for addressing the comments, LGTM!

Issue #741, opened 2026-08-03
Add topic flow library


nadahlberg:
- library/courts/<court>/topics/<topic>/config.yml + flows/*.yml + forms/*.pdf.
- Topic-level apply (topic fields + all its flows) and single-flow apply.
- Upsert semantics: fields match by name so stored answers and deadline calendar UIDs survive a re-apply.
- Add the check-added-large-files exclusion for library form PDFs.
- Second library panel in the admin shell.

PR #754, opened 2026-08-05
Make Site a singleton model

### LINKED ISSUE #731: Enforce Site as a singleton model

nadahlberg:
Fixes #731

This PR does a few things:
- Enforces the existence of exactly one row for the Site model, also removes `name` and `active` from Site (since there's only one now)
- Adds a migration to handle deletion of extra old site rows
- Adds a signal to ensure an initial site row exists (this logic is removed from the `seed_data` command, as this command will be deleted once the topic library #741 lands)
- Removes UI components and endpoints related to toggling the active site.
- Simplified site-related selectors and services to those that deal with a single site.
- Drops the `site` field from the Topic model. All related selectors and services are no longer scoped by site.
- Makes a `cache` module as a place to store cache keys, used by `topic_list` and `site_get`. And adds a decorator pattern for to the topic and site services that can be used to wrap services that modify those objects and therefor need to bust the cache.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -164,7 +164,9 @@ The compact rules (board mechanics live at the org level; sizing history, anchor
 
 **Naming:** services and selectors are `{model_name}_{action}` — `site_get_active`, `topic_create`, `user_identity_merge`, `user_upload_llm_parts`. Allow some liberty when a utility genuinely implicates two models equally. Anything not part of a module's public surface takes a leading underscore.
 
-**Reads are selectors, writes are services.** `user_list` reads, so it's a selector; `user_developer_toggle` writes, so it's a service. A cache key lives in the selector module that reads through it, and whoever invalidates it imports it from there.
+**Reads are selectors, writes are services.** `user_list` reads, so it's a selector; `user_developer_toggle` writes, so it's a service.
+
+**Cross-layer constants get their own module**, so no domain module has to import a constant out of another one: cache keys in `app/cache.py`, group and permission names in `app/permissions.py`. Each is read, written, and invalidated from a different layer, and anchoring one to any single consumer makes the other two reach sideways for it.
 
 **Permissions.** Admin access is the `app.manage_site` and `app.manage_developers` permissions, carried by the `Admins` and `Developers` groups (provisioned by a `post_migrate` receiver in `signals.py`). Check them with `request.user.has_perm(...)` in views and `{% if perms.app.manage_site %}` in templates — don't wrap either in a selector. Page views use Django's `@permission_required(..., raise_exception=True)`; the JSON API keeps its own `manage_site_required` / `manage_developers_required` decorators, because the built-in renders an HTML 403 where those endpoints must return a JSON body. These are deliberately separate from the auto-generated `add/change/delete/view` permissions that gate Django admin. Full reference: [`docs/wiki/permissions.md`](docs/wiki/permissions.md).
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -168,7 +168,7 @@ The compact rules (board mechanics live at the org level; sizing history, anchor
 
 **Cross-layer constants get their own module**, so no domain module has to import a constant out of another one: cache keys in `app/cache.py`, group and permission names in `app/permissions.py`. Each is read, written, and invalidated from a different layer, and anchoring one to any single consumer makes the other two reach sideways for it.
 
-**Cached reads are invalidated at commit, never before.** A selector that caches wears a `timeout=None`, and every service that writes the same rows wears `@busts(KEY)` from `services/utils.py`. Deferring the delete to commit is the whole point: an immediate delete lets a reader on another connection repopulate the key with the pre-commit value, and nothing busts it a second time, so the stale copy is permanent. The same rule binds management commands — `transaction.on_commit`, not a bare `cache.delete`. Cached values are pickled model instances, so the keys carry a version suffix to bump whenever the model's fields change.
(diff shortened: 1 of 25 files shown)


CLAUDE.md
```diff
@@ -162,9 +162,13 @@ The compact rules (board mechanics live at the org level; sizing history, anchor
...
 
 **Permissions.** Admin access is the `app.manage_site` and `app.manage_developers` permissions, carried by the `Admins` and `Developers` groups (provisioned by a `post_migrate` receiver in `signals.py`). Check them with `request.user.has_perm(...)` in views and `{% if perms.app.manage_site %}` in templates — don't wrap either in a selector. Page views use Django's `@permission_required(..., raise_exception=True)`; the JSON API keeps its own `manage_site_required` / `manage_developers_required` decorators, because the built-in renders an HTML 403 where those endpoints must return a JSON body. These are deliberately separate from the auto-generated `add/change/delete/view` permissions that gate Django admin. Full reference: [`docs/wiki/permissions.md`](docs/wiki/permissions.md).
```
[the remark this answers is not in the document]
nadahlberg: Updated these references

litigant_portal/app/cache.py
```diff
@@ -0,0 +1,7 @@
+# The singleton Site row. Read through site_get, dropped on commit by
+# busts_site_cache, and cleared after migrate by ensure_site_row.
+SITE_CACHE_KEY = "site"
+
+# Every Topic in display order. Read through topic_list, dropped on commit
+# by busts_topic_list_cache.
+TOPIC_LIST_CACHE_KEY = "topic_list"
```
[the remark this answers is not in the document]
nadahlberg: Yes, not doing the version suffix since this is busted at migration time by ensure_site_row. I added busting the topic key here too so both get busted.

litigant_portal/app/models/site.py
```diff
@@ -5,13 +5,13 @@
 from .base import BaseModel
 from .choices import AI_MODEL_CHOICES, JurisdictionLevel, State
```
nicolearagao, line 7: nits related to commit messaging:
- b774b39 docs(claude): we aren't using a version suffix until we need it for cache keys (title not imperative)
- e3e8e03 feat(cache): fix busts_cache method reference on cache key comments (type mismatch, should be either docs or fix)
- 4721493 feat(conftest): clear test cache between tests (test instead of feat)

litigant_portal/app/migrations/0012_site_singleton.py
```diff
@@ -0,0 +1,121 @@
+# Hand-written: Site becomes a singleton (#731). makemigrations produces the
```
nicolearagao, line 1: Suggestion, not blocking: _prune_extra_sites and _pin_site_pk are the only new logic in this migration. test_site_singleton.py checks the result (one row, correct ID, constraint works) but does not test the migration itself,  for example, starting with two sites and topics that have the same slug in different sites.

Normally I would ask for this test, because the migration cannot be undone if it has a bug. But since the product has not launched yet, there is no real user data at risk. 

So this is optional. If you want the test anyway, to protect against future changes to this migration, you could write one using MigrationExecutor: migrate to 0011, create a stale site and a keeper site with topics that share a slug, migrate to 0012, then check that only one site remains with the correct ID.

litigant_portal/app/services/utils.py
```diff
@@ -0,0 +1,25 @@
+from functools import wraps
```
nicolearagao, line 1: Nice catch: busts_cache fixes a real race in the old code (cache.delete() right after save(), before commit).

Suggestion, not blocking: it's shared by two services now (site_update, topic_create/update/delete), but it's only tested indirectly through those callers. The one thing not covered anywhere: if the wrapped function raises, does the cache delete still get scheduled? Looking at the code it shouldn't (transaction.on_commit(...) is only reached after fn() returns), but that's true by reading the code, not by any test. Worth a small direct test for that case at some point.

litigant_portal/app/signals.py
```diff
@@ -2,15 +2,34 @@
...
 
+from .cache import SITE_CACHE_KEY, TOPIC_LIST_CACHE_KEY
+from .models import Site
 from .permissions import GROUP_PERMISSIONS
 from .services.user import user_identity_merge_anonymous
 
 logger = logging.getLogger(__name__)
 
 
+def ensure_site_row(sender, using=DEFAULT_DB_ALIAS, apps=None, **kwargs):
```
nicolearagao, line 17: ensure_site_row does two different things : creates the site row, and clears cache for both Site and Topic. But Topic has no link to Site anymore (this PR removes that link). If we add a third cached thing later, someone has to remember to come back to this function and add its key here. Easy to forget.

What do you think of putting a list like ALL_CACHE_KEYS = [SITE_CACHE_KEY, TOPIC_LIST_CACHE_KEY] in cache.py, and having this signal just clear that list? Then adding a new cache key later only means editing cache.py, where cache keys already live.

nicolearagao (dismissed):
LGTM, a few non-blocking suggestions

litigant_portal/app/models/site.py
nadahlberg: Good catches

litigant_portal/app/signals.py
nadahlberg: Added `DATA_MODEL_CACHE_KEYS` and `clear_data_model_cache` signal

nicolearagao (dismissed):
LGTM

nicolearagao (approved):

Issue #669, opened 2026-07-20
v2 chat: adopt site frame + view tests + token/contrast hygiene (parity phase 1)


mitch-mitchel:
Phase 1 of #668 — the smallest high-value slice, zero behavior risk: make v2 chat a portal page instead of a standalone app, without touching the template's structure.

### Problem

v2 chat renders outside the site frame (own minimal `v2/common/header.html`, not `c-organisms.header`), its view has zero tests and computes debug env for every request, and the template carries contrast fails (`greyscale-400` as text, 2.53:1) plus arbitrary values where theme tokens exist.

### What you'd like

- **Tests first** (`test_chat_v2_page.py`): 200 + template selection, `debug_env` in context for superusers only, anonymous access allowed
- **View**: gate `debug_env` on `request.user.is_superuser`; hoist the local `import os`
- **Frame**: drop the header block override (base default `c-organisms.header` renders); delete `v2/common/header.html`
- **Hygiene in place**: `text-greyscale-400` → `-500` where used as text (template + precomputed JS class strings); `min-h-[3.25rem]`→`min-h-13`, `min-h-[4.5rem]`→`min-h-18`; new `--text-2xs` theme token replacing the 9 `text-[9-11px]` arbitrary values
- **Zero-risk a11y**: `role="log" aria-live="polite"` on the messages container (SC 4.1.3); agent-state aside server-gated to superusers (matches the debug_env gate)

nicolearagao:
@mitch-mitchel Quick questions on phase 1:

- debug_env the issue says to gate it onis_superuser, but it doesn't exist anywhere in the codebase (greppedviews/,chat_view, everywhere). Nothing to gate yet. What should it actually contain settings.DEBUG, an env name, git sha, something else? I'll add it gated from the start rather than adding it ungated first.
- Header file the issue says deletev2/common/header.html, but the real file iscotton/organisms/chat_header.html, and it's shared withpages/admin/index.html. I can drop theheaderblock override inchat/index.htmlso base'sc-organisms.headerrenders there instead, but deleting the shared partial would break admin. Planning to just drop the override and leave the partial, flag if that's wrong.

mitch-mitchel:
@nicolearagao 

- 1st one might need a reassesment and it's own issue. I think how we handle debug is not solid yet
- the current chat header and header in the style guide are a merge conflict scenario. Will take some inspection to resolve witch bits combine into the final header. we can pair on it to discuss

Issue #605, opened 2026-07-06
a11y: shared quick wins (reduced-motion, external-link cues, contrast tokens)


mitch-mitchel:
## What

Small WCAG 2.2 AA fixes on **shared design-system** surfaces — the ones that aren't tied to any one page. Each is XS; grouped into one branch/PR.

- [ ] **Global reduced-motion guard** — `@media (prefers-reduced-motion: reduce)` block in `src/main.css` covering `typing-bounce`, `sidebar-flash`, `animate-spin`, `auto-dismiss`. *SC 2.3.3*
- [ ] **External-link cues** — route the `about` / `privacy` / `accessibility` content pages through the link atom's `sr-only` new-tab cue instead of hand-rolled `<a target="_blank">`. *SC 3.2.5*
- [ ] **Contrast: `greyscale-400` as text** — 2.53:1 on app background. Sweep text usages to `greyscale-500` (4.84:1), leaving decorative icon and border uses alone. *SC 1.4.3 Contrast*

## Why

WCAG AA is the org floor. These three are page-independent, so they hold regardless of which surfaces exist. Originally surfaced in the 2026-07-06 tech-debt WCAG spot-check; the #594/#595 a11y pass covered Topic Flow but not these shared tokens and animations.

**Scope change 2026-08-13.** This issue originally batched fixes against `pages/chat.html` and `chat.js`, both removed in the chat cutover (#682). Those findings — keyboard-operable action item, icon-only close button name, live region for streamed replies, deadline bell name — were never re-audited against the current chat, so they are **not** assumed to carry over. Chat-surface a11y belongs to #716; a fresh audit belongs there, not here.

## Definition of done

- Reduced-motion honored across all four animations
- External-link cues use the link atom on all three content pages
- No `greyscale-400` remains as a text color at under 4.5:1 contrast

mitch-mitchel:
**Re-scope (2026-07-20 tech-debt pass):** the surface this issue targets is shifting under it.

When this was filed, chat_v2 was deliberately excluded as unsettled UI. Since then #655 merged (7/14) and v2 is now the live development surface — and today we're rebuilding the v2 front end on the component library (site frame, cotton composition, v1-parity layout), with the deferred v2 a11y items built in rather than patched after: live region for streamed replies (SC 4.1.3 — v2 currently has none), thread-row keyboard operability, delete-modal accessible name + focus handling, icon-button names, and the greyscale-400-as-text contrast swaps.

What that leaves of this issue:

[...]

Issue #487, opened 2026-06-05
tech debt: input/select atoms double-emit attributes (explicit block + {{ attrs }})


mitch-mitchel:
### What needs to change

`c-atoms.input` and `c-atoms.select` emit every attribute **twice** — once through the explicit `{% if name %}…{% if required %}…` block, and again through the trailing catch-all `{{ attrs }}`. In the rendered tag, `name`/`id`/`value` appear twice and the `error` prop leaks as a literal, non-standard `error=""` attribute on the `<input>`:

```
<input type="text" name="n" id="t" … required class="…"  name="n" id="t" value="" … required error="">
                   └── explicit block ──┘                 └────── {{ attrs }} dump ──────┘
```

`c-atoms.checkbox` shares the same `{% if required %}` shape and the same latent trap.

### Why (motivation)

Surfaced during #486 review. This double-emission is what let an *optional* `fact_gather` field render HTML-`required` even after the molecule was corrected — `{{ attrs }}` re-added `required="False"`. We worked around it for `required` by consuming it in the atom `c-vars` (commit on `485-section-templates`), but the root pattern remains for `disabled`, `readonly`, and `error`.

- Duplicate attributes are messy; the leaked `error=""` is invalid HTML.
- Latent bug of the same class we just fixed: any prop passed to these atoms as a dynamic bool (e.g. `:disabled="False"`) stringifies to `"False"` (truthy) and/or leaks via `{{ attrs }}`, silently flipping a falsy flag on.

### Definition of done

- One source of truth per attribute: consume all declared props in `c-vars` (excluding them from `{{ attrs }}`), or drop the explicit block and rely solely on `{{ attrs }}`.
- `disabled`, `readonly`, `error` no longer duplicate or leak; `error` drives only the aria-invalid / state-class logic.
- `c-atoms.checkbox` reviewed and brought in line.
- Existing form-field tests stay green; add a focused assertion that a falsy `:disabled` produces no `disabled` attribute (mirrors the `required` test added in #486).

PR #778, opened 2026-08-06
refactor(cotton): extract the eyebrow section atom

### LINKED ISSUE #715: chat phase 2: extract c-atoms.eyebrow atom
[what]
Extract the uppercase section-label pattern (icon + small caps heading) that's currently hand-rolled 4x in pages/chat/index.html (Recent Activity x2, Briefcase heading) into a new `c-atoms.eyebrow` atom.

[why]
Part of the phase 2 cotton decomposition (#715, epic #668). Removes duplicated markup and converges on a single reusable atom.

[dod]
- New atom at cotton/atoms/eyebrow.html
- All 4 hand-rolled instances in the chat template replaced with `<c-atoms.eyebrow>`
- Style guide updated with the new atom

nicolearagao:
## What changed

Extracts the uppercase icon-plus-heading section label into a new c-atoms.eyebrow atom and converts the six hand-rolled instances: five in pages/chat/index.html (Recent Activity and Briefcase, each in the desktop aside and the below-lg drawer, plus the label-only "Your files" in the upload modal) and one in pages/admin/index.html.

The atom owns the icon and heading typography; callers pass their own spacing and borders via class. A level prop (2-6, default 2) keeps each call site at the right depth in the page outline: "Your files" sits under an h3, so it renders as h4.

**One visual change beyond the extraction. The admin panel header was text-greyscale-400, the same 2.53:1 contrast fail that #669 swapped out of the chat template. As greyscale-500 it measures 4.84:1 against the panel background, passing AA for 12px semibold.**

Closes [#715](https://github.com/freelawproject/litigant-portal/issues/715)

## Test plan

- [x] /style-guide/ renders the new section, all three variants, no console errors
- [x] Manual a11y pass: keyboard-only and 200% zoom

---


mitch-mitchel:
Clean extraction, and the `HandRolledEyebrowGuardTests` regex guard is the right test to have — it catches the failure mode extractions actually have.

Three notes, none blocking.

**1. `ChatPageEyebrowTests` uses `create_superuser` on a gate that moved.**

The Briefcase eyebrow is behind `{% if perms.app.manage_developers %}`. A superuser passes it because Django short-circuits `has_perm()` to `True`, so the assertion holds for the wrong reason — drop `manage_developers` from the `Developers` group and this test still goes green.

#753 merged today and moved `test_chat_page.py` off `create_superuser` onto `Group` membership for exactly this. This PR predates that, so it's not a regression — but update to match the group-membership pattern now that #753 has landed.

**2. The admin page is covered by a source scan, not a render.**

The module docstring says chat plus style guide "cover every prop," but the admin eyebrow (`wrench-screwdriver`, default level) only appears in `HandRolledEyebrowGuardTests`, which reads the template file rather than rendering it. A prop break on `pages/admin/index.html` would pass. Either render the admin page in a test or narrow the docstring claim.

**3. The admin heading color change is an a11y fix riding along.**

`greyscale-400` → `greyscale-500` on the admin panel eyebrow takes it from 2.53:1 to 4.84:1. Good change, but it's a behavior change inside a mechanical refactor, and it belongs to the same contrast sweep #605 now owns. Noting it there so it isn't done twice or reported as still outstanding.

Checked and cleared, for the record: `class="… {{ class }}" {{ attrs }}` matches `badge.html`, so it isn't #487's double-emit (that one is `input`/`select`, which carry explicit named-attribute blocks *as well as* `attrs`). And the `greyscale-400` icon is fine — `c-atoms.icon` sets `aria-hidden="true"`, and #605 exempts decorative icon uses.

mitch-mitchel (approved):
LGTM

Issue #724, opened 2026-07-31
Enable Dependabot security updates and scope the Actions token


mitch-mitchel:
### What needs to change?

Three repo settings:

- [ ] Enable Dependabot security updates (`dependabot_security_updates` is currently `disabled`)
- [ ] Set default workflow permissions to read (currently `write` repo-wide)
- [ ] Disable "Allow GitHub Actions to create and approve pull requests" (`can_approve_pull_request_reviews: true`)

### Why?

Dependabot alerts fire but produce no fix PRs, so clearing the 13 open alerts was hand work (#690 / #706).

Separately, `can_approve_pull_request_reviews: true` lets a workflow using `GITHUB_TOKEN` approve a PR, and the `main: review + CI` ruleset requires exactly 1 approval. That's a path around the review gate. Read-by-default token permissions are GitHub's own hardening guidance; grant write per-job where needed.

### Definition of done

- [ ] All three settings changed; no workflow breaks from the token scope-down

mitch-mitchel:
All three settings applied and verified:

- `automated-security-fixes`: enabled
- `default_workflow_permissions`: `read`
- `can_approve_pull_request_reviews`: `false`

No workflow breaks from the token scope-down. `stale.yml` already declares `issues`/`pull-requests: write`, and `deploy.yml`, `docker-build.yml`, `qa-deploy-do.yml`, and `staging-deploy.yml` each declare `contents: read`. `lint.yml` and `tests.yml` only checkout and run, so read is sufficient.

Notification routing was a separate gap, now fixed: org notifications route to a verified FLP address, per-event alert email on, weekly digest for standing alerts.

Four alerts remain open. #764 (aiohttp 3.14.3) clears three of them. The fourth, `cryptography`, is tracked separately since Dependabot would not have opened a PR for it.

Issue #136, opened 2026-02-25
SOC 2 compliance roadmap for court partner RFPs


mitch-mitchel:
## Context

Court partners will require compliance certifications in RFPs. SOC 2 Type II is the standard ask for SaaS vendors selling to state/local government. We need to understand the gap between where we are and what's required, and start closing it.

## Why SOC 2

| Framework | Fit | Notes |
|-----------|-----|-------|
| **[SOC 2 Type II](https://secureframe.com/hub/soc-2/trust-services-criteria)** | Primary target | Standard in court/gov SaaS RFPs. We choose applicable Trust Service Criteria. |
| **[StateRAMP](https://govramp.org/wp-content/uploads/2021/10/GOV_StateRAMP-Get-Started-Guide_FINAL.pdf)** | Next step | Increasingly required for state/local procurement ([28 states](https://www.carahsoft.com/wordpress/carahsoft-stateramp-the-importance-of-framework-harmonization-blog-2025/)). ~6 month certification. Built on NIST 800-53. |
| **[JTC/NCSC guidance](https://www.ncsc.org/about-us/committees/joint-technology-committee/jtc-court-technology-standards)** | Alignment signal | Court-specific [cybersecurity standards](https://www.ncsc.org/resources-courts/cybersecurity-basics-courts). Not a certification but referenced in RFPs. |
| HIPAA | Too strict | Only relevant if handling medical case data (guardianship health records). |
| [CJIS](https://www.nuharborsecurity.com/blog/cjis-compliance-requirements-the-2025-checklist-for-state-local-agencies) | Not applicable | Criminal justice information only. We handle civil self-rep litigant data. |
| [FedRAMP](https://www.fedramp.gov/) | Not applicable | Federal agencies only. Courts are state/local. |

See also: [FedRAMP vs SOC 2 vs CMMC vs StateRAMP comparison](https://fedrampgpt.com/blog/fedramp-vs-soc2-vs-cmmc-vs-stateramp/)

## Recommended Trust Service Criteria

SOC 2 lets you [choose which criteria](https://cloudsecurityalliance.org/blog/2023/10/05/the-5-soc-2-trust-services-criteria-explained) to include in your audit:

| Criteria | Include? | Rationale |
|----------|----------|-----------|
| **Security** | Yes (mandatory) | Access control, encryption, logging, incident response |
| **Privacy** | Yes | PII from litigants — collection, use, retention, disposal |
[...]

PR #779, opened 2026-08-07
ci(deploy): Updates pipeline to deploy `main` to production

ERosendo:
## What changed

Production deployments now run from `main`. A merge triggers the following pipeline:

```
tests -> build + push -> migrate/collectstatic -> rollout
```

The order is enforced through GitHub Actions job dependencies (`needs:`), so deployment only proceeds after the test suite succeeds.

The previous `staging` deployment pipeline has been removed. QA remains a manual deployment flow.

The action test is now exposed as a reusable workflow (`workflow_call`) and called by the deployment workflow. This keeps testing as part of the same pipeline instead of running as an independent workflow that could race the image publish step.


mitch-mitchel:
Nice shape — test gate via `needs:`, migrations in a throwaway pod before any prod pod is replaced, serialized concurrency. Four things.

**1. `setup-deploy`: both error handlers fire on any earlier failure.** `if: failure()` is job-scoped. Collectstatic fails → `Handle Migration Error` also runs and points at a pod to debug a migration that never ran. Needs step-scoping (`id:` + `steps.<id>.outcome == 'failure'`).

**2. Temp pod leaks on failure, and blocks the re-run.** `Delete Temporary Pod` has no `if:`, so it's skipped on failure. The header says the pod self-deletes after an hour, but `sleep 3600 & wait` under `--restart Never` leaves it Completed, not gone. Re-running the failed job then hits `AlreadyExists` on the same `temp-pod-<sha>`. Keeping it for debugging makes sense — delete-if-exists before `kubectl run`, or give it a TTL.

**3. `kubectl rollout status` has no `--timeout`.** Default waits forever, so a non-converging rollout pins the job to the 6h Actions limit. `--timeout=10m`.

**4. Static AWS keys.** `configure-aws-credentials@v5` does OIDC role assumption. Is a role available on the cluster side, or is that not stood up yet? Happy to take it as a follow-up if so — flagging because of #724/#725 and #136.

Also: no rollback path if `rollout status` fails, with migrations already applied. Backward-compatible migrations cover it and this PR keeps that rule, so I assume that's the v1 line — `kubectl rollout undo` if you want the cheap version.

Checked the required checks against the live ruleset: `lint`, `build (3.13)`, and `Docker build validation` are all `pull_request`-triggered, so dropping `push: main` from `tests.yml` doesn't weaken merge gating. DoD rewrite reads accurate.

ERosendo:
> setup-deploy: both error handlers fire on any earlier failure. if: failure() is job-scoped. Collectstatic fails → Handle Migration Error also runs and points at a pod to debug a migration that never ran. Needs step-scoping (id: + steps.<id>.outcome == 'failure').

Similar to the discussion here: https://github.com/freelawproject/litigant-portal/pull/598#discussion_r3515493179. I think the same reasoning applies in this case as well.

ERosendo:
> Temp pod leaks on failure, and blocks the re-run. Delete Temporary Pod has no if:, so it's skipped on failure. The header says the pod self-deletes after an hour, but sleep 3600 & wait under --restart Never leaves it Completed, not gone. Re-running the failed job then hits AlreadyExists on the same temp-pod-<sha>. Keeping it for debugging makes sense — delete-if-exists before kubectl run, or give it a TTL.

Addressed in [cf44146](https://github.com/freelawproject/litigant-portal/pull/779/commits/cf4414677df6301c12ed21a7d752444b7c3266dc).

This has only happened a few times for CL, and we have usually helped unblock the pipeline manually. I agree that it would be better if you could unblock it yourself by re-running the action.

ERosendo:
> kubectl rollout status has no --timeout. Default waits forever, so a non-converging rollout pins the job to the 6h Actions limit. --timeout=10m.

I agree that having some protection against a rollout hanging forever is useful, but I'm a little concerned about adding a fixed `--timeout=10m`.

CL usually has a large number of nodes where pods need to terminate gracefully, and the time required can vary depending on traffic conditions. In those cases, a healthy rollout could take longer than 10 minutes and we would incorrectly mark the deployment as failed.

ERosendo:
> configure-aws-credentials@v5 does OIDC role assumption. Is a role available on the cluster side, or is that not stood up yet? Happy to take it as a follow-up if so

I don't think a role is needed here. configure-aws-credentials supports multiple credential-loading methods, and OIDC is only one of them. OIDC is used when role-to-assume is provided, but this workflow is passing aws-access-key-id and aws-secret-access-key instead.

`staging-deploy.yml` used the same `@v5` configuration with static credentials to deploy the litigant namespace several times before. The setup-deploy step succeeded, which means aws eks update-kubeconfig was able to authenticate successfully 👍

ERosendo:
@mitch-mitchel Thanks! I addressed all four concerns you mentioned. It should be ready for another review whenever you have a chance.

mitch-mitchel (approved):
Approving. The temp-pod reap is a better fix than what I suggested — keeping the pod for inspection and sweeping by label at the start of the next deploy gets both properties, and the comment explaining why `Delete Temporary Pod` deliberately has no `if:` is the kind of thing that stops someone "fixing" it later.

On the error handlers: your call, and you're the one reading those logs at 2am. The mechanics are as described, but the cost is diagnostic noise on an already-failed deploy, and you've now said twice across two PRs that the action logs make the real failure obvious. That's operator ergonomics for the operator. Dropping it.

On OIDC: fair, I conflated "works" with "preferred". Static keys authenticate fine and match what CL already does everywhere, so it's not this PR's problem. If we ever want to move off long-lived keys it's an infra-wide decision, and #136 is the better home for it.

Two things I'd like to leave as notes rather than changes. **Please don't action either in this PR** so the approval stands:

1. **A rollout ceiling above your p99, rather than my arbitrary 10m.** Your point about graceful termination varying with traffic is right and I'd have created false failures. But with no `--timeout` at all, a genuinely stuck rollout burns up to the 6h job limit and holds the `litigant-prod-deploy` concurrency group the whole time, so the next merge queues behind a deploy that's already dead. Something like 45m gets the backstop without touching healthy rollouts — you know the real distribution, so pick the number.
2. **`skip-web-deploy` leaves the DB migrated ahead of the running web image.** That's safe under our backward-compatible-migrations rule and this PR keeps that rule, so it's consistent — just worth being explicit that the escape hatch depends on it.

Verified the required checks are unaffected: `lint`, `build (3.13)`, and `Docker build validation` are all `pull_request`-triggered, so dropping `push: main` from `tests.yml` doesn't weaken merge gating, and tests still run on main through the `workflow_call`. DoD rewrite reads accurate.

mitch-mitchel (approved):
Looks good!

nicolearagao (approved):

Issue #733, opened 2026-08-03
Contacts and resources on the settings tab


nadahlberg:
Adds the models and admin site UI for editing contacts and resources on the admin settings tab. 
- Add Contact and Resource models (unique name/label, order).
- CRUD + reorder endpoints; row_move helper in services/utils.py.
- Split the settings form into court details (explicit save) and AI models (save on change).
- Build out the settings tab UI for both lists.

PR #786, opened 2026-08-11
Add site and topic flow models

### LINKED ISSUE #736: Add topic flow models
- Add TopicFlow, TopicFlowFieldGroup, TopicFlowField, TopicFlowSection, TopicFlowLink, TopicFlowDeadline, TopicFlowForm, TopicFlowFormField, TopicFlowAnswer.
- One migration; additive only — the YAML corpus and app/topic_flow/ keep running untouched.
- No UI, no endpoints. Just the schema and its constraints.

nadahlberg:
## What changed

Fixes #736, Partially addresses #733

This PR adds the models needed for #733 (doesn't close the issue yet, as the settings page UI is needed to finish this). And the models for #736 (forgive me on the branch name).

### Site models

The two site models added are `Contacts` and `Resources`. These are both court-level resources that can be displayed to users, contacts includes contact info for people or entities, and resources are links to relevant help info.

#### Topic Flow
A `Topic` can now have multiple `TopicFlows` -- these represent more specific paths underneath the topic (e.g. eviction might have different flows for landlords vs tenants.

#### Text Content
The `TopicFlowSection` model represents the text content displayed to users. Each ordered section is composed of heading and content.

#### Links and deadlines
The `TopicFlowLink` and `TopicFlowDeadline` models just represent additional metadata. We should discuss whether we actually want to keep `TopicFlowLink` because for the current topic flows we've designed these are 1-1 with topic flow forms (so these might be better as a url field on forms, something to discuss). Deadlines represent important deadlines whose values can be populated by an offset relating to a topic flow field (discussed below).

#### Forms and FormFields

The `TopicFlowForm` and `TopicFlowFormField` models represent a relevant form or piece of paperwork associated with the flow. When the form is a PDF with AcroFields we extract these as `TopicFlowFormFields` -- we will be able to populate form fields from topic flow fields (details below).

#### Interview Structure

The `TopicFlowFieldGroup` and `TopicFlowField` models create the interview structure. A field group is just a way of grouping a set of fields together, representing a single page on the guided interview. The fields themselves should be the minimal set of fields needed to populate all of the `TopicFlowFormFields`

The relationship between flow fields and form fields need not be 1-1. We will implement a syntax for populating TopicFlowFormField.template so that form fields can be flexibly composed from the fields. For example, we might have topic flow fields for `first_name` and `last_name`, while a particular form might have a field for `full_name` whose template is `{first_name} {last_name}`.

Finally, we have `TopicFlowAnswer` which represents a user's value for a given flow field. The `TopicFlowAnswer.reviewed` field represents whether this was populated automatically by AI versus confirmed manually by the user.

### How these set us up for future features

- By pulling all of the `TopicFlowFields` we can dynamically generate a schema for the minimal set of fields needed for populating all of a topic flow's forms. This can be incorporated into the AI flow so that the model can pre-populate the values for these forms.
- The models for the interview structure are designed to include the necessary ingredients to generate a guided interview for rendering a dynamic UI so that the user can follow a guided interview to fill these fields out. They are also structured in a way that will make it possible to dump an appropriately formatted docassemble yaml.

## Test plan

---


litigant_portal/app/models/topic_flow.py
```diff
@@ -20,3 +20,205 @@ class Topic(BaseModel):
...
+    slug = models.SlugField(max_length=64)
+    name = models.CharField(max_length=255)
+    enabled = models.BooleanField(default=False)
+
+    class Meta:
+        ordering = ["created_at"]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["topic", "slug"], name="unique_topic_flow_slug"
+            )
```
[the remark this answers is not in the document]
nadahlberg: good idea, added

litigant_portal/app/models/topic_flow.py
```diff
@@ -20,3 +20,205 @@ class Topic(BaseModel):
 
     class Meta:
         ordering = ["order", "created_at"]
+
+
+class TopicFlow(BaseModel):
+    """A guided flow for a topic."""
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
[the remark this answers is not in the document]
nadahlberg: Leaving this for now. I opened up #787 to discuss if we want to change the base model id field. What you said is not quite true as `UserIdentity` currently inherits the BigAuto base field.

litigant_portal/app/models/topic_flow.py
```diff
@@ -20,3 +20,206 @@ class Topic(BaseModel):
...
+        NUMBER = "number", "Number"
+        CHOICE = "choice", "Choice"
+        BOOLEAN = "boolean", "Boolean"
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    group = models.ForeignKey(
+        TopicFlowFieldGroup,
+        on_delete=models.CASCADE,
+        related_name="fields",
+    )
```
nicolearagao, line 103: TopicFlowField.name needs a uniqueness constraint. Nothing stops two fields sharing a name:

TopicFlowField.objects.create(group=g, name="first_name")  # ok
TopicFlowField.objects.create(group=g, name="first_name")  # also ok

Two consequences already on this branch:

TopicFlowAnswer is unique on (identity, field), not (identity, name). So one litigant can hold two first_name answers, and nothing says which one is theirs.
The importer matches fields by name on re-import (the rationale at test_topic_flow_models.py:213). On a tie it updates whichever row the DB returns first, and the other row's answers are orphaned.
The YAML side already enforces this: loader.py:89-97 rejects duplicate question ids across the whole corpus.

Two options:

(group, name) is cheap, but only covers one page of the interview.
(flow, name) matches the loader scope, but needs a flow FK on Field, since a Postgres unique constraint can't reach across the group join (Django won't take group__flow in UniqueConstraint.fields either). The cost is that flow is then denormalized: move a group to another flow and the field's flow goes stale unless something keeps the two in sync.

Which do you prefer? Worth doing before merge either way. The tables are empty now, so it's one constraint. Later it needs a data migration to rename duplicates.

litigant_portal/app/models/topic_flow.py
```diff
@@ -20,3 +20,206 @@ class Topic(BaseModel):
...
+
+    class Meta:
+        ordering = ["order", "created_at"]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["group", "order"],
+                name="unique_group_field_order",
+                deferrable=models.Deferrable.DEFERRED,
+            )
+        ]
```
nicolearagao, line 122: Should this constraint apply to the other models too?
Eleven models on this branch have an order field: Topic, TopicFlow, TopicFlowSection, TopicFlowFieldGroup, TopicFlowField, TopicFlowLink, TopicFlowDeadline, TopicFlowForm, TopicFlowFormField, Contact, Resource. Only TopicFlowField has a unique constraint on it.

All eleven use ordering = ["order", "created_at"], so duplicates never crash, they fall back to insert time. The list stays stable, but the number the admin typed stops meaning anything. If the reorder helper only writes the row that moved, that shows up as: two sections both at order=0, the user drags one to the top, saves, and nothing changes on the page. A helper that renumbers every sibling hides it.

So is a duplicate order a bug or not? Three ways to answer, each with a real cost:

Add the constraint to the other ten. But every order field is default=0, so any two rows created without an explicit order would fail. That hits the importer, and tests like TopicFlowFieldGroup.objects.create(flow=self.flow).
Drop it from TopicFlowField, and make the reorder helper renumber siblings densely everywhere.
Keep it as is, and document why fields are the exception.
Worth deciding before the row_move helper in #733, since one helper serves all of them. For TopicFlowField it has to run inside a transaction, because the swap passes through a state the constraint forbids (see the test at test_topic_flow_models.py:98). For Contact it doesn't. If row_move is written and tested against Contact only, it will look fine and then fail on TopicFlowField.

litigant_portal/app/tests/test_topic_flow_models.py
```diff
@@ -0,0 +1,225 @@
...
+
+        with self.assertRaises(IntegrityError), transaction.atomic():
+            TopicFlow.objects.create(
+                topic=first.topic, slug=first.slug, name="Clash"
+            )
+
+    def test_flows_are_disabled_until_switched_on(self):
+        self.assertFalse(build_flow().enabled)
+
+    def test_deleting_a_topic_takes_its_flows(self):
```
nicolearagao, line 51: This covers topic→flow, and test_a_deadline_hangs_off_a_field_and_dies_with_it (line 143) covers field→deadline. But nothing calls flow.delete(), which is the deletion an admin will actually trigger.

It should take eight collections: sections, links, deadlines, field_groups and forms directly, then fields through group, answers through field, and mappings through form. The two deepest paths, answers and mappings, are the ones nothing exercises today.

litigant_portal/app/tests/test_topic_flow_models.py
```diff
@@ -0,0 +1,225 @@
...
+            TopicFlowDeadline.objects.filter(pk=deadline.pk).exists()
+        )
+
+
+@pytest.mark.postgres
+class FormMappingTests(TestCase):
+    def setUp(self):
+        self.flow = build_flow()
+
+    def test_form_slug_is_unique_per_flow(self):
```
nicolearagao: This only tests the clash, so it would still pass if the constraint were global instead of per-flow. test_slug_is_unique_per_topic_not_globally (line 35) gets this right: it first creates the same slug under a different topic to prove the scope. Same trick here, create a second flow and reuse slug="answer" on it.

nicolearagao (changes requested):
Schema looks solid, and the tests read well. One thing I'd like to settle before this merges: TopicFlowField.name has no uniqueness constraint (comment below). The tables are empty right now, so it's a one-line AddConstraint. After merge it needs a data migration to rename existing duplicates, which is why I'm flagging it now rather than as a follow-up. Either (group, name) or (flow, name) works for me, they just have different scopes and I'd like to know which you picked.

The rest is non-blocking. The order constraint question is worth an answer in the thread since it shapes the row_move helper in #733, but the change itself can land there. The two test comments are coverage on code that has no behavior yet, so they're nice to have.

litigant_portal/app/models/topic_flow.py
nadahlberg: Thanks! I think denormalizing makes sense in this case (and not just for the constraint's sake). In most cases we're just going to care about the whole bundle of flow fields. It's really only for rendering the guided interview where we care how they're grouped. So grouping really is a sort of merely-ancillary feature of fields.

litigant_portal/app/models/topic_flow.py
nadahlberg: Yeah the row_move helper is the place to enforce this. I think your second option makes sense -- drop the constraint from topic flow field.

nicolearagao (approved):
Thank you for addressing my comments! LGTM

PR #789, opened 2026-08-11
feat: audit transcript admin + retention guard

### LINKED ISSUE #745: audit logs: AI conversation output - define, implement, and verify retrievability
### What needs to change?

Before the September pilot, AI-generated conversation outputs must be written to an audit log and retrievable by FLP staff. This story covers both the design decision (what does a log entry look like, who can access it, how) and the implementation.

Open questions to resolve as part of this story:

Format: raw JSON message records, or a human-readable transcript export?
Access method: admin UI, management command, or direct DB query? Catie and Jessica need to be able to pull a transcript without dev involvement.
Identity: anonymous sessions have no user account — how is a session identified in a log for review purposes? (session key, thread ID?)
Scope: every AI message, or just assistant-role messages? Do tool calls get logged?
Retention: how long are logs kept before the cleanup job removes them? This must be decided before cleanup_sessions runs in production.

### Why?

The Definition of Done for the September pilot (issue #711) requires audit logs to be written and retrievable before launch. Without them, Catie cannot do the spot-check of AI conversation transcripts that is a sign-off gate, and there is no accountability trail if a litigant is harmed by an AI response.

Note: ChatMessage rows already exist in the database, so the underlying data is there. This story is about making it accessible and reviewable — not storing new data from scratch.

- [ ]The open questions above are answered and documented (a comment on this issue is enough)
- [ ]AI conversation transcripts are retrievable by an admin without writing custom queries — either via the Django admin, a management command, or an export endpoint
- [ ]At least one test transcript has been pulled end-to-end and reviewed to confirm it is complete and readable
- [ ]Retention behavior is explicit: the cleanup_sessions job does not delete transcripts that are within the audit retention window
- [ ]Catie or Jessica can pull a transcript themselves using documented steps before the go/no-go

nicolearagao:
Makes AI conversation transcripts retrievable by FLP staff before the September pilot, per the decisions posted on the issue:

- **Read-only "Chat threads" section in the Django admin.** Lists every AI conversation, searchable by user email, session key, thread description, or thread ID, filterable by date and thread type. The detail page shows the full transcript inline (user messages, AI answers, tool calls, tool results, and `[hidden]`/`[meta]` rows) with Markdown and JSON downloads. Any `is_staff` account can view; add/change/delete are blocked.
- **Export selectors** (`chat_thread_export_data`, `chat_thread_export_markdown`, `chat_thread_owner_label` in `selectors/chat_engine.py`), built from raw `ChatMessage.data` rather than the lossy frontend render path. JSON is the full-fidelity record (timestamps, token counts, cost, structured tool data); Markdown is the readable view.
- **Retention guard.** New `AUDIT_RETENTION_DAYS` setting (default 30, matching the command's previous default, so no privacy regression). `cleanup_sessions` never deletes an anonymous identity with chat activity inside the window, even when run with a smaller `--days`.
- **Staff doc**: `docs/wiki/audit-transcripts.md` covers pulling a transcript step by step, the retention policy, and known limits (user-initiated hard delete is deliberately deferred; prompts are not stored; session keys are lost at login merge).

## What changed

Part of # 745

## Test plan

- [ ] Catie or Jessica pulls one transcript end-to-end using the wiki steps (issue DoD, before go/no-go)
Screenshot From 2026-08-11 17-15-20
Screenshot From 2026-08-11 17-15-40

Json/Markdown files exports Examples
[transcript-2fa19002-9396-4237-8d89-bb6f51d7d5cf.json](https://github.com/user-attachments/files/30952457/transcript-2fa19002-9396-4237-8d89-bb6f51d7d5cf.json)
[transcript-2fa19002-9396-4237-8d89-bb6f51d7d5cf.md](https://github.com/user-attachments/files/30952458/transcript-2fa19002-9396-4237-8d89-bb6f51d7d5cf.md)

---

## Review notes

- Session keys of anonymous users appear in full in the admin and in downloaded exports; whether to truncate them is an open decision for the demo (see the known-limits comment on #745).
- User-initiated deletion still hard-deletes threads at any time; the guard only covers the cleanup job. Team decision pending, documented in the wiki page.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -27,6 +27,9 @@ OPENAI_API_KEY=your-openai-api-key-here
 # Chat model (LiteLLM format)
 # CHAT_MODEL=openai/gpt-4o-mini
 
+# Audit retention window in days — cleanup_sessions keeps chat activity this long
+# AUDIT_RETENTION_DAYS=30
+
 # PostgreSQL connection details
 # POSTGRES_HOST=localhost
 # POSTGRES_PORT=5432
--- a/docs/README.md
+++ b/docs/README.md
@@ -23,6 +23,7 @@ The repo's reference shelf: material you reach for when you need it. Day-to-day 
 | --------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
 | [SECURITY.md](./wiki/SECURITY.md)                   | Security architecture: production headers, secrets, CSP                                                           |
 | [ai-tone-guide.md](./wiki/ai-tone-guide.md)         | Tone and philosophy for AI-generated user-facing output                                                           |
+| [audit-transcripts.md](./wiki/audit-transcripts.md) | How staff pull AI conversation transcripts, plus the retention policy                                             |
 | [issue-conventions.md](./wiki/issue-conventions.md) | Issue templates, labels, and the reasoning behind them                                                            |
 | [permissions.md](./wiki/permissions.md)             | Admin panel permissions: the two permissions, the two groups, and how to check them                               |
 | [translation.md](./wiki/translation.md)             | gettext infrastructure reference (strategy: [#704](https://github.com/freelawproject/litigant-portal/issues/704)) |
--- a/docs/wiki/audit-transcripts.md
+++ b/docs/wiki/audit-transcripts.md
@@ -1,0 +1,34 @@
+# Pulling AI conversation transcripts (audit log)
+
+Who this is for: FLP staff who need to review AI conversations, for the pilot spot-check or after an incident. No database access or developer help is needed once your account is set up.
+
+## One-time setup
+
+You need a portal account with **staff status**. A developer grants this once (in the Django admin, check "Staff status" on your user). After that, everything below is self-service.
+
+## Pull a transcript
+
+1. Go to `/django-admin/` on the portal and log in.
+2. Click **Chat threads**. Every AI conversation on the site is listed here, newest first.
+3. Find the conversation you need. You can:
+   - **Search** by the user's email address, by session key (for anonymous users), by words in the thread description, or by pasting a full thread ID.
+   - **Filter** by date or thread type using the sidebar.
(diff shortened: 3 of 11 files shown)


nadahlberg (dismissed):
LGTM

Two things to get your thoughts on:
1. Do we want the retention window to be configurable by court admins? If so you could add a field to the `Site` model and use that instead of configuring `AUDIT_RETENTION_DAYS` from the env.
2. Is it worth keeping the `--days` arg in the `cleanup_sessions` script? I'm imagining that in production we'll have a cron job that runs this daily or so. Would there be any reason to retain more anon data in the db beyond the audit retention period?

nicolearagao:
> LGTM
> 
> Two things to get your thoughts on:
> 
>     1. Do we want the retention window to be configurable by court admins? If so you could add a field to the `Site` model and use that instead of configuring `AUDIT_RETENTION_DAYS` from the env.
> 
>     2. Is it worth keeping the `--days` arg in the `cleanup_sessions` script? I'm imagining that in production we'll have a cron job that runs this daily or so. Would there be any reason to retain more anon data in the db beyond the audit retention period?

1. I'd say keep it in env for this one. Site is a singleton, so env vars are already per-court configuration. The bigger issue I am thinking about is what kind of setting this is. Site holds court-facing presentation and model choice: things a court admin should tune freely and that are reversible. Retention length decides how long transcripts sit in the database. Putting it behind manage_site means anyone with that permission can shorten it in a form field and have the next cron run destroy records irreversibly, with no review and no deploy trail. The whole point of the guard in cleanup_sessions is to stop accidental destruction of audit data, and it gets a lot weaker if the value it reads is editable by the same person it's guarding against.
Where I'd change my mind: if we ever host multiple courts in one deployment, or a court's legal team needs to lengthen retention without waiting on an FLP deploy.  I don't think we need it yet. 

2. No, nothing needs keeping past the audit window, so the cron is just cleanup_sessions --delete with no args (the default is already AUDIT_RETENTION_DAYS).
I'd still keep the arg though, because --days gates on identity age while the audit guard gates on chat activity. A smaller --days cleans up anonymous identities that never chatted, which have no audit value, and a larger one is useful for a one-off backlog purge before the cron goes live. Neither is something the daily job does.

nadahlberg (approved):
@nicolearagao makes sense

PR #810, opened 2026-08-19
Rebuild topic flow models: global variable glossary and conditional form packets

### LINKED ISSUE #809: Global variable glossary and conditional form packets for topic flows

Interview fields and forms are owned by individual topic flows, which costs us in two ways.

Shared facts are re-declared and re-asked per flow. A litigant's name, address, county, and date of birth are separate field rows in every flow that needs them, and an answer given in one topic cannot prefill another. The ND name-change tracks already duplicate roughly 90% of their field sets to differ in 10%.

A flow's form packet is fixed. There is no way to vary which forms a litigant needs based on what we learn about them: the fee-waiver forms when they cannot pay the filing fee, the criminal-history disclosure and good-faith statement when a felony conviction applies, the money-claim addendum when a landlord is also claiming back rent. Today the only expression of branching is cloning near-identical sibling flows.

Restructure the topic flow schema around two app-wide pools plus one linking model, so flows compose shared assets instead of owning copies.

**`Variable` / `VariableAnswer`:** one glossary of atomic facts, each named once app-wide. `is_global` marks the handful the assistant tracks from the start of any conversation; everything else comes into play once a packet is known. Answers are per-identity and follow the person across topics. A variable can gate another (`asked_when`), so follow-up questions only surface when relevant.

**`Form` / `FormField`:** fillable PDFs as shared assets belonging to no flow. A form field's template composes glossary variables, so the variables a form needs are derived from its templates rather than declared.

**`TopicFlowFormCondition`:** the only edge between a flow and a form. With no variable the form is always in the flow's packet; with one, it is in the packet only while the identity's answer matches. Branching is not a separate concept: a "branch condition" is an ordinary glossary variable that some condition row points at, and the role is derived, never declared on the variable.

Interview paging (`TopicFlowInterviewPage` / `TopicFlowInterviewVariable`) is retained as purely cosmetic grouping for guided-interview export.

This mirrors how docassemble resolves interviews (a document you never assemble never gets its questions asked), so conditional packets and conditional questions remain one mechanism end to end.

nadahlberg:
Closes #809.

Reshapes the topic flow schema so flows compose two app-wide pools instead of owning
per-flow copies of fields and forms. `Topic`, `TopicFlow`, `TopicFlowSection`, and
`TopicFlowLink` are untouched.

| Before | After |
| --- | --- |
| `TopicFlowField` (flow-scoped) | `Variable` — global, `name` unique app-wide, `is_global`, `question`, `asked_when` gate |
| `TopicFlowAnswer` | `VariableAnswer` — `unique(identity, variable)`, follows the person across topics |
| `TopicFlowForm` (owned the PDF) | `Form` — global slug + file |
| `TopicFlowFormField` | `FormField` — same columns, hangs off `Form` |
| — | `TopicFlowFormCondition` — flow, form, optional (variable, operator, value) |
| `TopicFlowFieldGroup` | `TopicFlowInterviewPage` + `TopicFlowInterviewVariable` (cosmetic paging) |
| `TopicFlowDeadline` | kept, `offset_from` repointed at `Variable` |

Decisions worth review:

- **PROTECT** where deleting a variable would silently change what litigants are told
  to file or when: `TopicFlowFormCondition.variable`, `TopicFlowDeadline.offset_from`,
  and the `asked_when` self-FK. Interview placements stay CASCADE (presentation only).
- **Two check constraints**: a condition `value` requires a `variable`, and
  `asked_when_value` requires `asked_when` — a half-written condition cannot read as
  conditional while behaving as unconditional.
- **No unique on `(flow, form)`**: repeat condition rows are allowed and OR together,
  the escape hatch for a form that more than one answer pulls in.

Tacked on, same theme — choices convention cleanup:

- New convention: model choices always live in `models/choices.py`. Moved
  `TopicFlowFormConditionOperator` and `VariableDataType` there.
- Consolidated the duplicate US state lists onto `choices.State` (canonical ordering:
  DC filed alphabetically, territories appended — the old profile copy dangled DC last
  and lacked the territories). Labels are now `gettext_lazy`-wrapped: the es catalog
  already carried all 51 state names as msgids from the old profile list, so unwrapped
  labels would have shrunk the catalog on the next `makemessages`.
- `JurisdictionLevel` left untranslated deliberately: admin-facing only,
  auto-generated labels.

Migration `0014` is plain delete-and-create with no data handling — aside from `Topic`
(unchanged), none of these tables were populated. Verified applied against the dev
Postgres container: all seven new tables present, four old tables dropped, Topic rows
intact. Includes the `userprofile.state` AlterField from the choices consolidation.

Tests: `test_topic_flow_models.py` rewritten against the new schema (27 tests). The
interesting ones are the deliberate inversions — variable and form uniqueness went
from per-flow to app-wide, and the deadline test asserts its anchor is *protected*
where the old one asserted cascade. The flagship,
`test_deleting_a_flow_takes_its_composition_and_spares_the_pools`, pins the whole
architecture: a flow's deletion reaches all six flow-scoped collections and leaves the
glossary, answers, and forms standing.

Out of scope, next PR: porting the `wip` surfaces (selectors/services/views/agent
tools and the library YAML importer) onto this schema.


litigant_portal/app/models/topic_flow.py
```diff
@@ -62,168 +63,221 @@ class Meta:
...
-    """A page of fields within a flow's interview."""
+class Variable(BaseModel):
+    """A fact about the person or their case, named once app-wide."""
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-    flow = models.ForeignKey(
-        TopicFlow,
-        on_delete=models.CASCADE,
-        related_name="field_groups",
+    name = models.CharField(max_length=255, unique=True)
```
nicolearagao: where does Variable.name validation live? Names become docassemble variable names, so they must be snake_case, and names are permanent API. This PR leaves name a bare CharField(max_length=255, unique=True) with no enforcement anywhere yet. Sould we enforce this in the YAML loader checks (next PR), on the model, or both?

litigant_portal/app/models/topic_flow.py
```diff
@@ -62,168 +63,221 @@ class Meta:
...
 
     class Meta:
         ordering = ["order", "created_at"]
+        constraints = [
+            models.CheckConstraint(
+                condition=(
+                    models.Q(variable__isnull=False)
+                    | models.Q(value__isnull=True)
+                ),
+                name="form_condition_value_requires_variable",
```
nicolearagao, line 216: The form_condition_value_requires_variable constraint blocks value-without-variable, but the mirror shape is still legal: variable set, value NULL. What should the evaluator do with (variable=can_pay_filing_fee, operator=equals, value=NULL)?

nicolearagao (commented):
It's looking really good, I have two questions.

litigant_portal/app/models/topic_flow.py
nadahlberg: - **Enforce at YAML checks:** Yes, added it to the list. 
- **Enforce at model-write-level:** We could enforce it when writing any services like `variable_create` or `variable_update`. 
- **Enforce in the model definition itself**: Is that possible? If so, I like it (and makes sense to do in this PR)

nicolearagao: yes, it's possible. You can add a CheckConstraint with name__regex=r"^[a-z][a-z0-9_]*$" in Variable.Meta. Then Postgres enforces it on every write, same style as the two constraints you already have. I think it fits in this PR.
I would also add a RegexValidator on the field, so admin/forms show a normal error message instead of an IntegrityError.

litigant_portal/app/models/topic_flow.py
nadahlberg: Should that be legal? I was thinking that `null` would be the value when the VariableAnswer hasn't been populated yet for the user. E.g. `can_pay_filing_fee` would be a condition for a `Request for Fee Waiver` only when we know for a fact that the user can't afford to pay i.e. is `False`

Wyt?

litigant_portal/app/models/topic_flow.py
nadahlberg: Very cool! Added both

litigant_portal/app/models/topic_flow.py
nicolearagao: Agree, it should not be legal. If NULL means "no answer yet", then a condition can never match on NULL, so the row shape makes no sense.

So I would add the mirror constraint in this PR: variable and value must be both set or both null.

```
models.CheckConstraint(
    condition=(
        models.Q(variable__isnull=True)
        | models.Q(value__isnull=False)
    ),
    name="form_condition_variable_requires_value",
)
```

Together with your existing constraint, the only legal shapes are "no condition" and "complete condition". Same thing for asked_when / asked_when_value on Variable.

nadahlberg: Ohh, I see! Done.

nicolearagao (approved):

Issue #790, opened 2026-08-11
Store prompts with chat threads so audits show the AI's exact instructions


nicolearagao:
### What problem does this solve?

Audit transcripts (#745) include every persisted message, but the system prompt and tool-internal prompts are never stored. They live in code, are rebuilt on every turn (`agent.generate_system_prompt`, prepended in `_messages_for_llm`), and can change mid-conversation (`refresh_system_prompt`). A transcript shows what the AI said, but not what it was told to do at the time. Prompts change with releases, so for an old conversation there is no way to reconstruct the instructions behind a response. This is the main gap in the audit record, documented as a known limit in docs/wiki/audit-transcripts.md and in the decisions comment on #745.

### What would you like to see?

Persist prompts alongside the thread so the audit export is a full record:

- Capture the rendered system prompt at the moment it is used: on a thread's first turn, and again whenever it is regenerated after `refresh_system_prompt`. Tool-internal prompts (thread-description prompt, QueryDocument's prompt) are candidates too.
- Prompts repeat heavily across turns and threads. Store deduplicated artifacts (content hash to prompt text) and reference the artifact from the thread or message, instead of copying the full text into every row.
- Surface them in the #745 exports: the JSON download includes the prompt text or references; the Markdown transcript shows the prompt where it first applies and wherever it changes.
- Retention follows the thread: prompt references die with the thread under the existing audit window; shared artifacts are garbage-collected when nothing references them.

Open questions for the story:
- Storage shape: hidden/meta ChatMessage rows vs a separate prompt-artifact model.
- Whether to also record the deployed code version (git SHA) per turn as a cheaper complement.

nicolearagao:
Design decided (Slack, 2026-08-12, with @Mitch and @Nathan Dahlberg). This settles the two open questions in the description.

Storage shape: a separate model, not hidden/meta ChatMessage rows. One row per distinct instruction state, holding the rendered system prompt text, the serialized tool schemas, and a content hash of both together. The row is exactly what we sent the model.

Linking: ChatMessage gets a foreign key to that row. Every message points at the state in force when it was produced, so a mid-thread refresh_system_prompt is recorded with no extra work. This is better than referencing from the thread, which cannot show per-message changes.

Dedup: get_or_create on the content hash. This dedups across all threads, not only inside one. It also handles the case where a refresh regenerates an identical prompt, which can happen because check_weather sets refresh_system_prompt=True. Query cost per turn is the same as comparing against the previous message.

[...]

PR #811, opened 2026-08-19
Record deployed git SHA per assistant turn

### LINKED ISSUE #801: Record the deployed git SHA per assistant turn so old chats can be reconstituted

Audit transcripts (#745) show what the AI said, but not what it was told to do. The system prompt and tool-internal prompts are rebuilt from code on every turn and are never stored, so once a release changes the prompts, the instructions behind an old conversation can't be reconstructed. This is documented as the main gap in the audit record (docs/wiki/audit-transcripts.md).

Storing full prompt texts as deduplicated artifacts is the long-term fix, tracked in #790. We agreed that is post-pilot work. For September we need a cheaper way to reconstitute a chat if an audit requires it.

Phase 1 of #790: record the deployed git SHA with each assistant turn.

- Expose the running commit to Django (for example a GIT_SHA env var baked into the image at build time; "unknown" as the local-dev fallback).
- Stamp the SHA on each assistant message (or on the thread, re-stamped when the value changes, since a deploy can happen mid-conversation).
- Surface it in the #745 exports: JSON includes the SHA per turn; the Markdown transcript shows it once at the start and again wherever it changes.
- Document the reconstitution procedure in docs/wiki/audit-transcripts.md: check out the SHA, read the prompt layers under litigant_portal/prompts/, and combine with the transcript.

Known limit to state in the docs: the SHA reconstructs the code-side prompt layers only. Prompt content that comes from the database (Site admin config, corpus edits) is not versioned by the SHA. If any of that materially shapes the prompt, note it in the reconstitution doc; fully solving it is #790's job.

Out of scope (stays in #790): storing rendered prompt texts, deduplicated prompt artifacts, retention/garbage collection of prompt records.

_No response_

nicolearagao:
## What changed

Phase 1 of #790: stamps the deployed commit SHA on every chat message so an old conversation can be reconstituted against the exact code that produced it, without storing full prompt text.

- `settings.GIT_SHA` reads the `GIT_SHA` env var (default `"unknown"` for local dev); wired into the Dockerfile build arg, `make build-image`, and the QA deploy workflow.
- `ChatMessage.git_sha` is stamped from `settings.GIT_SHA` in `chat_message_create`, so every message (not just assistant turns) carries the SHA of the code that produced it — including across a mid-conversation deploy.
- The audit exports (#745) surface it: JSON carries `git_sha` per message; Markdown shows it once in the header and flags a `**Deployed SHA changed to ...**` line wherever it changes.
- `docs/wiki/audit-transcripts.md` documents the reconstitution procedure — check out the SHA, then read the actual live prompt sources (`BASE_PROMPT` in `agents/assistant.py` and tool prompts like `agents/tools/query_document.py`), not the dormant `litigant_portal/prompts/` scaffolding — plus the known limit that database-sourced prompt content isn't covered by the SHA.

Closes #801

## What changed

Phase 1 of #790: stamps the deployed commit SHA on every chat message so an old conversation can be reconstituted against the exact code that produced it, without storing full prompt text.

- `settings.GIT_SHA` reads the `GIT_SHA` env var (default `"unknown"` for local dev); wired into the Dockerfile build arg, `make build-image`, and the QA deploy workflow.
- `ChatMessage.git_sha` is stamped from `settings.GIT_SHA` in `chat_message_create`, so every message (not just assistant turns) carries the SHA of the code that produced it — including across a mid-conversation deploy.
- The audit exports (#745) surface it: JSON carries `git_sha` per message; Markdown shows it once in the header and flags a `**Deployed SHA changed to ...**` line wherever it changes.
- `docs/wiki/audit-transcripts.md` documents the reconstitution procedure — check out the SHA, then read the actual live prompt sources (`BASE_PROMPT` in `agents/assistant.py` and tool prompts like `agents/tools/query_document.py`), not the dormant `litigant_portal/prompts/` scaffolding — plus the known limit that database-sourced prompt content isn't covered by the SHA.

Closes #801

## Test plan

- [x] Manual: ran a chat turn locally, confirmed `git_sha` on the created rows and on both transcript downloads (`unknown`, as expected with no `GIT_SHA` set in dev)
- [x] Manual: built the image with `--build-arg GIT_SHA=<sha>` and confirmed it lands as an env var in the container

---


nadahlberg (approved):
LGTM. Hoping we never have to use this 😬

Issue #726, opened 2026-07-31
Pin GitHub Actions to commit SHAs


mitch-mitchel:
### What needs to change?

- [ ] Pin floating third-party tags: `astral-sh/setup-uv@v6`, `pre-commit/action@v3.0.1`, `aws-actions/configure-aws-credentials@v5`
- [ ] Reconcile `actions/checkout` — currently both `@v4` and `@v6` in the same repo
- [ ] Consider enabling `sha_pinning_required` once pinned

### Why?

A floating tag can be repointed to different code by whoever controls the action.

`appleboy/ssh-action` (SSHes into the QA box) is already SHA-pinned, so the highest-risk one is covered. This closes the rest.

Sized S rather than XS because pinning needs a pass to verify nothing breaks, plus a plan for keeping pins current.

### Definition of done

- [ ] All third-party actions SHA-pinned with a version comment
- [ ] CI green

nicolearagao:
deploy.yml still needs SHAs

PR #812, opened 2026-08-20
ci: pin github actions to shas and configure dependabot

nicolearagao:
## What changed

- **Pin and bump are separate commits.** `9e46a16` moves no versions at all: each SHA is the exact commit the floating tag already resolved to, so behavior is unchanged by construction. `e1583cb` then takes the upgrades. If CI had broken, the cause was attributable and the bump revertable without giving up the pins.
- **Upgrade only where a PR proves it.** `lint.yml` / `tests.yml` run on `pull_request`, so checkout v7 and setup-python v7 are exercised here. Everywhere else is pin-only, and Dependabot proposes upgrades in a reviewable PR instead.
- **`setup-uv` stays at v6.8.0.** v6 to v10 is four majors, and upstream stopped publishing major tags at v8 (citing tj-actions, same reasoning as this PR). Wants its own PR.
- **`actions/stale` stays at v9.1.0** though v11.0.0 exists — nothing here can exercise it.
- **`checkout` reconcile is deliberately incomplete.** v7.0.1 in CI, v6 in `deploy.yml`,
  v4.3.1 in the two Docker workflows. Finishing it would bump paths this PR can't test.
  Dependabot closes the gap.

Refs #726 

## Not in this PR

**`deploy.yml` — six floating actions, deliberately untouched.** It runs on neither
`pull_request` nor `push`, so nothing here could verify it, and it holds the AWS
credentials. It needs its own PR once we know how a deploy change gets exercised before
merge. #726 stays open for this.

**Enabling `sha_pinning_required`** — a setting, not a file, so it can't live in a commit.
See the sequencing note below.

## Test plan

CI green on this PR:

| Check | Result |
| --- | --- |
| `lint` | pass, 17s |
| `build (3.13)` | pass, 1m21s |
| Docker build validation | pass, 39s |

That covers all pins in `lint.yml` and `tests.yml` at their new versions. Before pushing, each pin was checked twice: the SHA exists upstream, **and** it matches the SHA of the version in its trailing comment. The second check matters more — a correct SHA with a lying comment survives review.

`stale.yml` is not exercised by CI (no matching trigger). It'll be confirmed by the next scheduled Monday 14:00 UTC run.

## Follow-ups

1. **`deploy.yml` pinning** — blocked on the deploy-verification question.
2. **Enable `sha_pinning_required`** — must come **after** (1), not after this PR. Turning it on while `deploy.yml` still has floating tags would fail the next deploy.
---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+
+updates:
+  # Version updates for GitHub Actions. Security updates already run without this
+  # file (they're driven by advisories + the dependency graph); version updates are
+  # what keep SHA-pinned actions from going stale.
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+    # One grouped PR per week instead of one per action.
+    groups:
+      github-actions:
+        patterns:
+          - '*'
+    commit-message:
+      prefix: build(deps)
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,12 +10,12 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
 
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
         with:
           python-version: '3.13'
 
-      - uses: pre-commit/action@v3.0.1
+      - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
         env:
           SKIP: uv-lock-check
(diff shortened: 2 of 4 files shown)


nicolearagao (commented):
@mitch-mitchel no rush, this is for when you're back from PTO.

This PR leaves deploy.yml unpinned on purpose: it runs on neither pull_request nor push, so nothing in a PR can prove the pins work before merge, and it holds the AWS credentials. Before I do the follow-up PR, I need to know how a deploy.yml change gets exercised before it merges. 
@ERosendo tagging you too since this touches the deploy pipeline. You probably know best how deploy changes are usually verified.

Related: once deploy.yml is pinned, we want to enable sha_pinning_required on the repo (in that order, enabling it first would fail the next deploy).

mitch-mitchel:
this makes sense to me, you can convert it to ready

nicolearagao:
> this makes sense to me, you can convert it to ready

Undrafting this. Items below should be addressed in a follow-up PR:

deploy.yml — six floating actions, deliberately untouched. It runs on neither
pull_request nor push, so nothing here could verify it, and it holds the AWS
credentials. It needs its own PR once we know how a deploy change gets exercised before
merge. https://github.com/freelawproject/litigant-portal/issues/726 stays open for this.

Enabling sha_pinning_required — a setting, not a file, so it can't live in a commit.
See the sequencing note below.

mitch-mitchel (approved):
👍

PR #814, opened 2026-08-23
Corpus pipeline: YAML content tree, validation, and database sync

### LINKED ISSUE #813: Corpus pipeline: YAML content tree, validation, and database sync
### What needs to change?

Add a court content corpus and the pipeline that validates it and keeps the database in sync with it.

- `litigant_portal/corpus/`: the content tree. `variables.yml` (app-wide variable glossary, 31 variables), `forms/` (one yml per Form plus its fillable PDF, 8 forms / 6 PDFs including the ND fee waiver packet), and `courts/<court>/topics/<topic>/flows/<flow>.yml` for North Dakota adult name change (standard + waiver) and Franklin County OH eviction (tenant + landlord). Identity comes from paths; conventions documented in the corpus README.
- `selectors/corpus.py`: the corpus contract. Pydantic schemas mirror the Django models they populate (field for field, so sync can unpack `model_dump()` directly), loaders parse the tree, and `CorpusSchema` cross-validates everything (PDF AcroForm field membership, template variable resolution, interview coverage, gate and condition ordering, orphan detection). Fail-fast: first problem raises.
- `checks/corpus.py`: thin Django system check adapter surfacing an invalid corpus as `corpus.E001`. The checks module is now a package split by domain (`corpus.py`, `prompts.py`).
- `services/corpus.py`: `corpus_sync(court=None, strict=False)`. Upserts by natural key (variable name, form/topic/flow slug) so stored `VariableAnswer` rows survive corpus edits; flow composition rows replace wholesale. `strict` deletes rows the corpus no longer names, except variables, which are only flagged `in_schema=False` (answers cascade with their variable, so orphans wait for a curated migration instead). `court` narrows topics, flows, and site config to one court and falls back to the new `CORPUS_COURT` setting; unset syncs every court's topics, flows, contacts, and resources and leaves the site's court fields alone.
- `sync_corpus` management command wraps the service and replaces `seed_data` in the docker entrypoint (runs `--strict` on boot, so the database always mirrors the corpus).
- Migration 0015: `Variable.default` becomes a typed JSONField (matching gate/condition values) and adds `Variable.in_schema`.
- `make check` target for running system checks in the container.

### Why?

The beta demo needs real per-court content (interviews, packets, deadlines) behind the topic flow models that landed in #810, and partners need an explicit validated contract for authoring it. This realizes the planned library domain under the corpus name and is foundation for deleting the legacy `app/topic_flow` engine and its flat `content/*.yml` corpus wholesale.

- `manage.py check` validates the corpus and fails red on any malformed or unrecognized content
- `sync_corpus` is idempotent, and strict mode reconciles renames without losing stored answers
- Fresh environments boot with the corpus imported (entrypoint), scoped by `CORPUS_COURT` when set
- Bench confirmation of the ND declaration conviction-table row mappings is tracked separately

nadahlberg:
Closes #813

This PR does three things:

1. Builds out a file structure in `corpus` for the hard-coded court and topic flow configurations.
2. Adds a `check_corpus` check with extensive validation.
3. Adds a `sync_corpus` command that replaces `seed_data` and upserts updated corpus data into the DB.

A few additional changes along the way:
- `checks.py` is converted to a directory with the new checks in `checks/corpus.py`
- Added a `make check` convenience command to the makefile to run `manage check` in the docker container.
- Two tweaks to the `Variable` model: the `default` field type (made it consistent with other `VariableValue` type fields and an added `in_schema` field.

## 1. The corpus directory structure

The new structure consolidates variables, forms, court configs, and topic flows in a format that closely mirrors our data models:

```
litigant_portal/corpus/
├── README.md
├── variables.yml                        # app-wide glossary: every Variable, by name
├── forms/                               # shared Form pool: <slug>.yml + its PDF
│   ├── nd-fee-waiver-petition.{yml,pdf}
│   ├── nd-name-change-petition.yml         ┐ two forms sharing
│   ├── nd-name-change-petition-waiver.yml  ┘ nd-name-change-petition.pdf
│   ├── nd-name-change-declaration.yml         ┐ two forms sharing
│   ├── nd-name-change-declaration-waiver.yml  ┘ nd-name-change-declaration.pdf
│   ├── nd-name-change-notice.{yml,pdf}
│   ├── nd-name-change-confidential-info.{yml,pdf}
│   └── nd-name-change-order.{yml,pdf}
└── courts/
    ├── north-dakota/
    │   ├── court.yml                    # Site court fields + contacts + resources
    │   └── topics/adult-name-change/
    │       ├── topic.yml
    │       └── flows/{standard,waiver}.yml
    └── franklin-county-oh/
        ├── court.yml
        └── topics/eviction/
            ├── topic.yml
            └── flows/{tenant,landlord}.yml
```

## 2. Corpus validation

The corpus is validated by `checks/corpus.py::check_corpus`. The logic for this check lives in two places:

- `checks/corpus.py::check_corpus_directory_structure`: Checks for the directory structure itself.
- `selectors/corpus.py`: Validates the corpus data itself with pydantic models.

The `selectors/corpus.py` is a big file, but ultimately boils down to several Pydantic models with their validation baked in and a few loader methods for loading models from hardcoded corpus files. 

All of this boils down to a single `corpus_load` method which loads a single `CorpusSchema` model that consolidates all of the parts:

```
class CorpusSchema(BaseSchema):
    variables: dict[VariableNameField, VariableSchema]
    forms: dict[SlugField, FormSchema]
    form_acro_fields: dict[str, set[str]]
    courts: dict[SlugField, CourtSchema]
    topics: dict[tuple[SlugField, SlugField], TopicSchema]
    flows: dict[tuple[SlugField, SlugField, SlugField], FlowSchema]
```

Validation that crosses document type is applied at the `CorpusSchema` model. All validators include docstrings that list the rules they test in bullet points.

## 3. Corpus sync command

A new management command `sync_corpus` is added, the logic lives in `services/corpus.py`. A few details:

- Data is upserted so that we aren't deleting rows with each update.
- `--strict` mode (which is used by the entrypoint) deletes rows that aren't in sync
- `Variable` rows are never deleted (even in `strict`) because these link to `VariableAnswers` which would wipe user data. Instead an `in_schema` field is added to variables and set false if the var is orphaned. E.g. if we change a variable name, then the old var will be flagged and we can write a migration to cleanly migrate the user answers without losing them.

The `CORPUS_COURT` decides which court config and topic flows are imported. When set, this is straightforward (we import the corpus data relevant to that court only). When unset, which we'll use for dev, the following happens:

`CORPUS_COURT=None` (dev mode):
- All topic flows across all courts are imported
- Court metadata we add to `Site` model is ignored, not synced
- `Contacts` and `Resources` across all courts are imported. The reason this will be ok, is that we planned on linking specific subsets of contacts and resources to topic flows anyway, which is what will render on the public-facing views. So once #815 is implemented, the "dev" versions of topic flow pages will mirror the court versions of those pages anyway.


litigant_portal/app/services/corpus.py
```diff
@@ -0,0 +1,258 @@
...
+    for name, schema in corpus.variables.items():
+        row = rows.get(name) or Variable(name=name)
+        _apply(row, schema, exclude={"asked_when"})
+        row.in_schema = True
+        row.save()
+        rows[name] = row
+    for name, schema in corpus.variables.items():
+        row = rows[name]
+        gate = schema.asked_when
+        row.asked_when = rows[gate.variable] if gate else None
```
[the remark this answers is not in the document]
nadahlberg: Good catch. I don't like the having multiple forms for a single PDF when that PDF can be filled out differently. Let's just duplicate the PDF in this rare scenario so that forms and PDFs remain 1-1. Will resolve once applied.

nadahlberg:
@nicolearagao lmk what you think about the Claudebot's migration note. We don't have any data in the variables table, so another one of those 'do we do a custom migration for good practice purposes' scenarios.

litigant_portal/app/selectors/corpus.py
```diff
@@ -0,0 +1,634 @@
...
+
+# Forms
+
+
+class FormFieldMappingSchema(BaseSchema):
+    """Populates a ``FormField`` row."""
+
+    pdf_field: str = Field(min_length=1)
+    template: str = ""
+    checked_when: str = ""
```
nicolearagao: checked_when is the only schema field with no validation, and a typo here silently prints a wrong court form.

Right now nothing catches:

- a value that isn't one of the variable's choices ('citzen' = never-checked box);
- mixed boolean spellings (the ND forms use both 'True' and 'yes');
- what checked_when: '' means (equality would check the box; the YAML comments say it stays blank).
- Can we validate checked_when against the variable's legal values, and write down the rule for the empty case?

docker/django/entrypoint.sh
```diff
@@ -14,9 +14,9 @@ run_bootstrap_superuser() {
     manage bootstrap_superuser
 }
 
-run_seed_data() {
-    echo "Seeding baseline data ..."
-    manage seed_data
+run_sync_corpus() {
+    echo "Syncing corpus ..."
+    manage sync_corpus --strict
```
nicolearagao, line 19: Boot runs --strict in both dev and prod, so every restart deletes any Topic, Flow, Form, Contact, or Resource the corpus doesn't name, and re-overwrites the site's court fields. Nothing else writes contacts/resources today, so this is coherent, but it means anything an admin creates in those tables dies at the next restart.

Is the corpus now the sole owner of these tables, including admin edits? Fine if yes, I just want to understand.

litigant_portal/corpus/courts/franklin-county-oh/topics/eviction/flows/landlord.yml
```diff
@@ -0,0 +1,67 @@
+# Eviction (Franklin County, OH), LANDLORD role. The directory locates the
```
nicolearagao: Tests: missing

The v1 engine this replaces has test_topic_flow_schema.py, test_topic_flow_loader.py, and test_topic_flow_validation.py. The v2 equivalents are missing, and this layer is the partner-facing contract, so the rules need regression protection. What I'd like covered:

- Schema validators : each validator docstring already lists its rules ("choice values unique", "no asked_when cycles", "pdf_field exists in the PDF", "gate placed before gated", "consumed variable on a page"...). One failing fixture per rule.

corpus_sync:

- run it twice: same row counts, same primary keys, no suffixed files in storage (the bug 3eafaba fixed has no regression test);
- --strict deletes stale topics/flows/forms/contacts/resources but only flags variables
CORPUS_COURT set vs unset: court filtering and whether site fields are written;
a VariableAnswer (with its reviewed flag) survives a re-sync. This is the trust boundary, one test.
- System check: one test that the shipped corpus passes corpus_load() (right now CI never validates the real corpus, only Docker boot does), and one that a broken fixture corpus returns corpus.E001.

Practical suggestion: a tmp MEDIA_ROOT and a tiny one-field fixture PDF, so the sync tests don't churn the real 1.5 MB PDFs.

nicolearagao (changes requested):
Overall this looks good. The schema validation is thorough, the sync is careful about identity (natural keys, answers surviving edits), and the corpus content itself is excellent. 

The one thing I'm holding the review on is tests. This PR adds ~900 lines of Python that will carry real court forms, and none of it is covered. The v1 engine this replaces has test_topic_flow_schema.py, test_topic_flow_loader.py, and test_topic_flow_validation.py, so this is the bar the repo already set, and this layer is the partner-facing contract: the rules need regression protection before people start authoring against them. What I'd like covered:

Schema validators: each validator docstring already lists its rules ("choice values unique", "no asked_when cycles", "pdf_field exists in the PDF", "gate placed before gated", "consumed variable on a page"...). One failing fixture per rule.

corpus_sync:

- run it twice: same row counts, same primary keys, no suffixed files in storage (the bug 3eafaba fixed has no regression test);
- --strict deletes stale topics/flows/forms/contacts/resources but only flags variables;
- CORPUS_COURT set vs unset: court filtering and whether site fields are written;
a VariableAnswer (with its reviewed flag) survives a re-sync. This is the trust boundary, one test.
- System check: one test that the shipped corpus passes corpus_load() (right now CI never validates the real corpus, only Docker boot does), and one that a broken fixture corpus returns corpus.E001.

Practical suggestion: a tmp MEDIA_ROOT and a tiny one-field fixture PDF, so the sync tests don't churn the real 1.5 MB PDFs.

One more thing: a 0015 migration just merged to main, so this branch's 0015_variable_default_and_in_schema now collides with it. Please pull main and renumber yours to 0016 (with its dependencies pointing at the new 0015) before the next push, so we don't merge a duplicate migration number.

update: 0016 is in main already. rebase and check latest migration =)

litigant_portal/app/selectors/corpus.py
nadahlberg: Ok I'm reworking the checked_when system. Seeing a lot of problems in addition to the ones you flagged:
- The "`checked_when` is the value and `template` is the variable" pattern is hacky. It makes `template` take on two meanings.
- It also doesn't give us a clean FK for the field referenced in the template
- It doesn't mirror our other "when" fields which use `*_when` to reference the variable and `*_when_ value` for the value

I think we can do this:
- `checked_when` (FK to the field) and `checked_when_value` JSON value expression
- Docassemble also lets checkboxes be always checked/unchecked. So add one more `checked` field that defaults to None and where True /  False correspond to the always states.
- Add constraints so that: (i) `checked_when` and `checked_when_value` must be null or not null together, (ii) `checked` and `checked_when` are mutually exclusive, (iii) `template` must be null if either `checked` or `checked_when` are set

In short, FormFields would populate PDFs in exactly one of three ways:
- The `template`
- Always on or off checkboxes with `checked`
- Conditionally checked by another variable via `checked_when` + `checked_when_value`

nadahlberg: Really on the fence about this. I'm going to commit it, but idk. Wyt?

docker/django/entrypoint.sh
nadahlberg: I think that's the intent for now. The admin site is too incomplete to expose to anyone, so I'm fine with overwriting the few editable fields for now if you are.

litigant_portal/app/selectors/corpus.py
nicolearagao: Hey, the FK redesign looks good to me. It fixes the real problem, template was doing two jobs before and now it isn't. And it matches how we already do this for Variable.asked_when and TopicFlowFormCondition. So it's not too much, it fits what's already there.

One thing is still missing though: the new shape doesn't stop a typo like 'citzen' or 'yes' instead of true from silently checking the wrong box. That was my original worry.

Good timing: I just built variable_value_validate in services/topic_flow.py for #821. It checks a raw value against a variable's type and choices. If your corpus loader uses this same function for checked_when_value, it catches bad choices and mixed true/false spellings for free. And it means we only have one place that checks values, which is what the issue asked for.

One more small ask: can you add one line saying what an empty checked_when_value means? That was my third question and it's still open.

nicolearagao (approved):
LGTM

Issue #580, opened 2026-06-29
Add litigant portal assistant


nadahlberg:
Add a new agent in the new chat framework for the user-facing LP assistant. Should reproduce the components of the old one:

- System prompt loader method (global instructions + topic data from state + user info + case index)
- State definition (active topic)
- A tool for collecting case information

Issue #177, opened 2026-03-17
Briefcase: unified session state model for user data, chat, and case context


mitch-mitchel:
## Summary

Introduce the "briefcase" as the single source of truth for all user session state. Every interaction (chat, topic entry, search, user input) writes to the briefcase. Every view (chat, dashboard, sidebar, detail pages) reads from it and filters for what's most useful and actionable.

## Mental Model

The briefcase is a virtual container holding everything the portal knows about a user's situation:

- **Chat histories** — full conversation logs across sessions
- **Legal issue data** — extracted facts, defenses, case type
- **Dates & deadlines** — court dates, filing deadlines, program application windows
- **People & places** — court addresses, opposing parties, household info
- **Action items** — next steps surfaced by AI or selected by user
- **Documents & resources** — search results, saved references, uploaded files
- **Eligibility & programs** — benefits the user qualifies for (e.g., ILRPP)

## How It Works

- **Writes are additive** — chat extracts facts, topic cards pre-populate context, search saves results, user provides details. All routes feed the same briefcase.
- **Views filter by context** — the chat page shows conversation + relevant sidebar items. A dashboard shows prioritized deadlines and action items. Detail views let users dig deeper. The briefcase holds everything; the view layer decides what's most useful right now.
- **Origin doesn't matter** — whether the user arrived via topic card, search bar, or deep link, the briefcase accumulates the same way. No routing logic based on entry point.

## Future Vision

The briefcase becomes the app-wide context layer:

- **API integrations** — court data, legal aid availability, e-filing status all write to the briefcase
- **AI tooling** — agents read briefcase context to give better answers, write discoveries back
- **Legal aid collaboration** — attorneys or legal aid workers helping a user see the same briefcase, filtered for their role
- **All architecture is abstracted** — to the user (and helpers), API calls, AI reasoning, and data sources are invisible. The briefcase just has their stuff.

## Relationship to Existing Work

- **#172** (backend data layer) remains separate — models, AI agent, form generation are implementation concerns the briefcase will consume
- **#173** (frontend topic flow) — current sidebar components (case_sidebar, deadline_card, action_item) become briefcase viewers
- **#174** (topic→chat routing, done) — established the entry points that will write to the briefcase

## Open Questions

- Briefcase persistence: session-scoped (anonymous) vs. account-linked (authenticated)?
- Granularity of briefcase items — flat list vs. grouped by legal issue?
- How does the briefcase handle multiple legal issues (eviction + child support)?

[...]

PR #816, opened 2026-08-23
feat(ai): store prompt artifacts for auditable chat exports

### LINKED ISSUE #825: Store the rendered system prompt and tool schemas per assistant turn
### What needs to change?

The prompt-artifact slice of #790: persist the exact instruction state behind each model-backed assistant message, and surface it in the JSON audit export.

- A `PromptArtifact` model holding the rendered system prompt, the tool-schema snapshot, and a SHA-256 content hash, deduplicated so repeated instruction states share one row.
- A nullable reference from `ChatMessage`, populated for model-backed assistant messages including tool-call-only turns, left null for user, tool, hidden, meta and legacy messages.
- JSON audit export carries a per-message reference plus a deduplicated top-level `prompt_artifacts` collection.
- `docs/wiki/audit-transcripts.md` updated to describe what the export now contains and what it still does not.

### Why?

#790 is the parent and covers more than this: tool-internal prompts, Markdown transcript rendering, and garbage collection of unreferenced artifacts. Splitting the slice out gives the work that is actually in flight an issue it can close, and leaves #790 open for the rest rather than closing it early.

- [ ] System prompt and tool-schema snapshot persisted per model-backed assistant turn
- [ ] Identical instruction states share one row; a changed prompt or changed schema produces a new one
- [ ] JSON export defines each referenced artifact once and handles legacy null references
- [ ] Audit-transcript docs state the remaining limits

Explicitly out of scope, tracked on #790: tool-internal prompt capture, Markdown transcript rendering, orphan cleanup and retention, model-name tracking.

zeitcow:
## What changed

- Add a `PromptArtifact` model to store the rendered system prompt, tool-schema snapshot, and unique SHA-256 content hash.
- Add a nullable, protected `ChatMessage.prompt_artifact` relationship.
- Capture the exact `generate_system_prompt()` value and the same `agent.tool_schemas` snapshot used for each model call.
- Link model-backed assistant messages, including tool-call-only responses, to the prompt artifact used for that turn.
- Leave user, tool, hidden, meta, and legacy messages unlinked.
- Extend JSON audit exports with nullable per-message references and a deduplicated top-level `prompt_artifacts` collection.
- Preserve the deployed `git_sha` behavior that landed separately in #811 while integrating prompt-artifact references into the same chat-message and audit-export paths.
- Update the audit-transcript documentation to describe JSON prompt artifacts, deployed-SHA provenance, legacy behavior, and the remaining Markdown/tool-internal prompt limits.
- Add coverage for hashing, deduplication, refreshed prompts, no-tools calls, message-link behavior, legacy/export behavior, and compatibility with the deployed-SHA changes.

This keeps the scope to the prompt-artifact slice of #790. It does not add Markdown prompt rendering, orphan cleanup, tool-internal prompt capture, model-name tracking, or deployed git SHA tracking; the SHA work landed separately in #811 / #801.

Closes #825

Part of #790

## Test plan

Verified against the PostgreSQL-backed Docker environment after rebasing onto current `main`:

- [x] `manage makemigrations --check --dry-run`, no changes detected
- [x] Prompt-artifact, git-SHA, and settings tests, 13 passed
- [x] Audit export and audit-admin tests, 33 passed
- [x] Related prompt/chat/message/upload/cleanup tests, 110 passed
- [x] Full test suite, 529 passed, 18 subtests passed
- [x] Fresh PostgreSQL migration through both 0014 branches, upstream 0015 merge, and prompt-artifact 0016
- [x] Ruff format check for changed Python files
- [x] Ruff check passed
- [x] `git diff --check`
- [x] `make pre-commit` equivalent green on Windows, all configured hooks plus full Python 3.13/PostgreSQL suite

## AI Disclosure

- [ ] No AI tools were used to create the content of this PR.
- [x] Parts of this PR were created with the help of an AI tool, and I have carefully reviewed all of its content and take full responsibility for it.

AI model used:

- ChatGPT Sol 5.6 via Codex

I reviewed the code, tests, migration graph, audit export behavior, documentation updates, PostgreSQL test results, and final diff, and take responsibility for the contribution.


mitch-mitchel:
@claude review

zeitcow:
Rebased onto current main after #811 and resolved the overlapping chat-message/audit-export changes. Migration graph is clean, and the full PostgreSQL suite passes (529 tests, 18 subtests). Scope remains limited to the prompt-artifact/JSON-export slice of #790.

mitch-mitchel:
Read through this. Nice work — the scoping is honest, the tests are real, and the PR body made review easy.

The tests earn a specific callout: `test_hash_preserves_schema_array_order` catches the subtle failure (`sort_keys=True` orders object keys, but schema list order is meaningful), and `RefreshingAgent` exercises a mid-loop change rather than just the happy path. `select_related("prompt_artifact")` in the export keeps it off N+1.

One thing to record rather than fix here. `on_delete=PROTECT` on a hash-deduplicated, globally shared artifact is the right call for an audit record, and it is safe today because `LitigantAssistant.generate_system_prompt()` returns a static `BASE_PROMPT` and ignores `thread_id`. Two things sit against that:

- #790 asks for the opposite lifecycle — "shared artifacts are garbage-collected when nothing references them." PROTECT blocks that, so the follow-up will need to reconcile the two.
- `thread_id` is in the signature because prompts are meant to become per-thread. #179 (inject court terms and corpus on topic entry), #177 (briefcase session state) and #580 (prompt loader, state definition) each make that true. When one lands, dedup stops collapsing and every row holds one litigant's injected case context, in a shared table that cannot be deleted.

Neither blocks this slice. Capturing it so it is a decision later rather than a surprise.

One housekeeping ask: you were right not to write `Closes #790` — this does not close it. We have split the slice out as #825, so if you add `Closes #825` to the description, the issue moves and closes on merge on its own. Nothing else needed from you.

zeitcow:
Thanks for the detailed review. I’ve added Closes #825 to the description. The PROTECT/lifecycle and future per-thread prompt considerations make sense to carry into the follow-up work.

mitch-mitchel (approved):
Looks good, let's :shipit:

PR #817, opened 2026-08-24
fix(migrations): merge the two 0014 leaf nodes

nicolearagao:
## What changed

migration numbered 0014 against 0013 as its parent. Git merged them cleanly (different new files, no textual overlap), but the app was left with two leaf nodes in the migration graph. `migrate` refuses to run at all in that state, so every DB-backed test errored during `django_db_setup` rather than failing on its own terms.

Empty merge migration joining the chain back to a single leaf. Nothing to reconcile in `operations`: one 0014 adds `ChatMessage.git_sha`, the other creates the Form/Variable tables, so the two touch disjoint models. A database that already applied either one picks up the other plus this no-op, so no local or production reset is needed.

Guards to catch the next collision before it reaches main are tracked separately.

## Test plan

---

### DIFF
--- a/litigant_portal/app/migrations/0015_merge_20260824_1200.py
+++ b/litigant_portal/app/migrations/0015_merge_20260824_1200.py
@@ -1,0 +1,10 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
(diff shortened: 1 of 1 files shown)


mitch-mitchel (approved):
Looks Good

PR #820, opened 2026-08-24
docs: document runnable DB-free fast tests

### LINKED ISSUE #480: docs(claude): carve out the runnable DB-free fast suite from the no-run-tests note
## What

Refine the `make lint` / `make test` / `make pre-commit` note in `CLAUDE.md` (Commands section). It currently says sandbox restrictions prevent Claude from running tests and to "never attempt to run them yourself" — but that's only true for the `make` / Docker / Postgres targets. The **DB-free fast suite runs directly** via `.tox/fast/bin/pytest <file>` (when the `.tox/fast` env is already built), which lets the agent do a real RED→GREEN TDD cycle for non-DB units.

## Why

The blanket "never run tests yourself" led to authoring tests + implementation in the same step and declaring green without ever observing the test fail — i.e. skipping the RED step that makes TDD real. Carving out the runnable fast suite restores genuine TDD (write test → run → confirm it fails for the right reason → implement → green) for DB-free units like `topic_flow/` (renderer, AnswerStore, deadlines, artifacts).

## Definition of done

- [x] Update the note to distinguish: **can't** run `make` / Docker / Postgres-marked targets; **can** run DB-free fast tests directly via `.tox/fast/bin/pytest <path>` (when `.tox/fast` exists) for the RED→GREEN cycle
- [x] State that Postgres-marked / view tests still need `make test` / Docker (Mitch runs those)
- [x] Preserve the existing guidance: Mitch runs the full lint/test workflow; a single post-commit mention, no re-prompting
- [x] Keep it consistent with the global CLAUDE.md TDD/mutation testing practice

HarperZ9:
## What changed

- distinguish Docker and Postgres-backed `make` targets from the runnable DB-free fast suite
- document `.tox/fast/bin/pytest <path>` for a focused RED to GREEN cycle when the environment already exists
- clarify that the fast marker filter can still select unmarked tests that use Django's database

Closes #480

## Test plan

- [x] `uv run pre-commit run --all-files`
- [x] `uv run pre-commit run prettier --files CLAUDE.md`
- [x] `uv run pre-commit run --files CLAUDE.md` after the requested-change patch
- [x] `git diff --check origin/main...HEAD`
- [ ] `make pre-commit` green (GNU Make and Docker are unavailable on this Windows host; the draft PR's CI must run the full suite)

`uv run tox -e fast -- -q --tb=short` was also started, but it produced no terminal result after more than five minutes and was stopped. This pull request changes repository guidance only.

### AI assistance disclosure

I used OpenAI Codex with `5.6 Sol` (`gpt-5.6-sol`) to help inspect the public issue and repository guidance and to draft this small documentation change. I reviewed and understand every changed sentence and take responsibility for the contribution. No nonpublic personal information, sealed or restricted record, litigant data, credential, or private repository material was provided to the model. The AI service was used through a logged-in account with model-improvement data sharing disabled before this work.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -27,7 +27,7 @@ Chat model is configurable via `CHAT_MODEL` env var (LiteLLM format, e.g. `opena
 
 **`make lint`, `make test`, `make pre-commit`** — sandbox restrictions prevent Claude from running these Docker- and Postgres-backed targets. Mitch runs the full lint/test workflow as part of his own process. A single post-commit mention is plenty; don't re-prompt about it across the commit/PR steps.
 
-**DB-free fast tests** — when `.tox/fast` already exists, Claude can run a focused test directly with `.tox/fast/bin/pytest <path>`. Use that suite for a real RED→GREEN cycle on non-DB units: write the focused test, run it and confirm that it fails for the expected reason, implement the change, then rerun it to green. Postgres-marked tests and view tests still require `make test` and Docker, so Mitch runs those with the full workflow.
+**DB-free fast tests** — when `.tox/fast` already exists, Claude can run a focused test directly with `.tox/fast/bin/pytest <path>`. Use that suite for a real RED→GREEN cycle on non-DB units: write the focused test, run it and confirm that it fails for the expected reason, implement the change, then rerun it to green. Tests marked `postgres` require `make test` and Docker. The fast marker filter is not a complete database-isolation boundary: unmarked tests may still use Django's database, so run only focused tests already known to be DB-free through this path.
 
 ### Local Development (Docker)
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -25,7 +25,9 @@ Chat model is configurable via `CHAT_MODEL` env var (LiteLLM format, e.g. `opena
 
 ## Commands
 
-**`make lint`, `make test`, `make pre-commit`** — sandbox restrictions prevent Claude from running them. Never attempt to run them yourself. Mitch runs lint/test as part of his own workflow — a single post-commit mention is plenty; don't re-prompt about them across the commit/PR steps.
+**`make lint`, `make test`, `make pre-commit`** — sandbox restrictions prevent Claude from running these Docker- and Postgres-backed targets. Mitch runs the full lint/test workflow as part of his own process. A single post-commit mention is plenty; don't re-prompt about it across the commit/PR steps.
+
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -25,7 +25,9 @@ Chat model is configurable via `CHAT_MODEL` env var (LiteLLM format, e.g. `opena
 
 ## Commands
 
-**`make lint`, `make test`, `make pre-commit`** — sandbox restrictions prevent Claude from running them. Never attempt to run them yourself. Mitch runs lint/test as part of his own workflow — a single post-commit mention is plenty; don't re-prompt about them across the commit/PR steps.
+**`make lint`, `make test`, `make pre-commit`** — sandbox restrictions prevent Claude from running these Docker- and Postgres-backed targets. Mitch runs the full lint/test workflow as part of his own process. A single post-commit mention is plenty; don't re-prompt about it across the commit/PR steps.
```
nicolearagao, line 28: Thanks for taking the time to put this together, appreciate the contribution!

One process note for next time: could you comment on the issue asking to be assigned before starting work on it? That way we can avoid two people picking up the same issue in parallel, and it gives us a chance to flag any context you might need before you dive in.

On the PR itself: the DB-free fast-test workflow is a nice addition and matches the existing [testenv:fast] setup in tox.ini.

One inaccuracy to fix before merging: the new text says "Postgres-marked tests and view tests still require make test and Docker," implying view tests are excluded from the fast suite's -m "not postgres" filter. They aren't - test_topic_flow_view.py has 39 @pytest.mark.django_db tests with no postgres marker, so -m "not postgres" would still select them, and they'd fail on a DB connection error rather than being skipped.

Could you reword the CLAUDE.md line so it doesn't claim view tests are excluded from the fast path?
Otherwise this looks good to me

nicolearagao (changes requested):

CLAUDE.md
HarperZ9: Reworded the section to describe `postgres` marker handling and explicitly warn that unmarked tests may still access Django's database. Focused pre-commit checks for `CLAUDE.md` pass at a8bf3de.

HarperZ9:
Good call on the process note. I'll comment on the issue to get assigned before starting next time, so work doesn't get picked up twice.

The CLAUDE.md wording is fixed in a8bf3de: it no longer claims view tests are excluded from the fast path, and it warns that unmarked tests can still reach Django's database. I've marked the PR ready for review.

nicolearagao (approved):

PR #826, opened 2026-08-24
Guard against parallel migration collision

### LINKED ISSUE #818: Make parallel migrations fail loudly instead of breaking main
### What needs to change?

Two PRs branched from the same commit both added a `0014` migration (#810, #811). Git merged them cleanly because they were different new files, but the app ended up with two leaf nodes in the migration graph, so `migrate` refused to run and every DB-backed test errored at setup. Fixed after the fact by the merge migration `0015_merge_20260824_1200`.
Add guards so the next collision is caught before it lands on main:
- Add `django-linear-migrations` to `INSTALLED_APPS` and generate the per-app `max_migration.txt`. Two branches adding a migration then both edit that one line, so git raises a real text conflict at merge time.
- Add `python manage.py makemigrations --check --dry-run` to the lint or test workflow, so conflicting leaves (and models changed without a migration) surface as one clear line instead of hundreds of tracebacks.
Optional, worth discussing separately: turn on a merge queue for `main`. The ruleset currently has `strict_required_status_checks_policy: false`, so a PR tested against a stale `main` can merge into a `main` that has moved on. A merge queue tests `main + PR` before merging; plain strict checks would work too but force a manual rebase of every open PR after each merge.

### Why?

Green checks on a PR only prove the branch works against the commit it was branched from. Nothing ever tested the two branches combined, so the break only appeared on main, after the merge, as 354 setup errors with a single root cause. This will happen again whenever two migrations are in flight at once.

- [ ] `django-linear-migrations` installed and `max_migration.txt` committed and tracked.
- [ ] Two branches each adding a migration produce a git conflict on `max_migration.txt` (verify by hand).
- [ ] fails with a readable message when the migration graph has multiple leaves.

nicolearagao:
## What changed

- Add `django-linear-migrations` and commit the tracked `max_migration.txt`, so two branches adding a migration now conflict on that one line at merge time instead of silently landing two leaf nodes on `main` (how #810/#811 broke `migrate`).
- Add `makemigrations --check --dry-run` to the test run, so a broken migration graph fails with one readable line instead of hundreds of setup errors.
## Why
Green checks on a PR only prove the branch works against the commit it branched from. #810 and #811 both added a `0014` migration and both merged clean, but `main` ended up with two leaf nodes and every DB-backed test errored at setup. These two changes catch that class of collision before it lands, and fail loudly if it ever does.

### Why the package-data change is here
`makemigrations` runs Django's full system checks, and CI installs this
project as a real wheel — not editable. That wheel only ever included
`.py` files, so prompt schema JSON, templates, static assets, and locale
files were silently missing from it, and the checks that validate those
files (`chat.E001`/`E005`) failed. Docker never hit this (`pip install -e .`
reads straight from the source tree) and plain pytest never hit it either
(system checks only run via a management command). Adding the
migration-conflict check is what first built and ran against a real wheel,
exposing a gap that already existed. Fixing it here rather than splitting
it out because the new check can't pass CI without it — but it's a real
bug either way: any future switch of Docker to a non-editable install, or
packaging this for distribution, would have silently shipped without
these files.

Closes #818 

## Test plan

- [x] Simulated two branches each editing `max_migration.txt` and confirmed a real git merge conflict.
- [x] Simulated a two-leaf migration graph and confirmed `makemigrations --check` fails with `dlm.E005` (Conflicting migrations detected).
- [x] Confirmed the check passes cleanly (`No changes detected`, exit 0) on the current migration state.

---


nicolearagao (commented):
This PR's checks catch migration collisions specifically. The general case is still open: the branch ruleset has strict_required_status_checks_policy: false, meaning a PR is only tested against the commit it branched from, not the current tip of main - how #810/#811 both went green and still broke main.

Two options to close that gap:

- Flip strict_required_status_checks_policy to true. Simple, but open PRs need a manual rebase every time main moves.
- Merge queue for main. Tests main + PR automatically, no manual rebasing, but more setup and slightly slower merges.
@mitch curious which way you'd lean here, or if it's not worth it given how often we have concurrent PRs.

mitch-mitchel:
Good change, and the test plan actually simulated both failure modes rather than assuming them. One blocker and one question.

**Blocker: this branch is stale against `main` and would break the check it adds.** `max_migration.txt` here reads `0015_merge_20260824_1200`, but `main`'s leaf is now `0016_promptartifact_chatmessage_prompt_artifact` (#816, merged 23:15, an hour after this opened). All four checks are green because they ran against this branch's base. Merged as-is, `makemigrations --check` fails on `main` immediately, because the recorded max no longer names the real leaf.

Which is #818's own thesis playing out on #818's own fix: green checks only prove the branch works against the commit it branched from.

- [x] Rebase onto `main` and update `max_migration.txt` to `0016_promptartifact_chatmessage_prompt_artifact`

**Question: the `[tool.setuptools.package-data]` block.** Fourteen lines covering templates, static, locale, content and prompts, with nothing in the description about packaging. If adding an INSTALLED_APP surfaced a real packaging gap under tox, that is worth a line in the body so the next person knows why it is there. If it is unrelated, it belongs in its own PR rather than riding a P0.

- [x] Explain it in the description, or split it out

**Not blocking, but this is the second stale-base incident today** — the first put 354 errors on `main`. #818 parked the merge queue as "optional, worth discussing separately." Two occurrences in one day across three active authors reads more like evidence than theory. Worth its own issue rather than expanding this one.

mitch-mitchel:
Re: the automation vs manual questions - I prefer automation even if it's slower ci/cd. We can optimize later if it gets really slow @nicolearagao

nicolearagao:
> Re: the automation vs manual questions - I prefer automation even if it's slower ci/cd. We can optimize later if it gets really slow @nicolearagao

@mitch-mitchel rebased and fixed the migration file. Added on the PR description why the package-data change was necessary, commit also explains it. 
Created follow-up issue https://github.com/freelawproject/litigant-portal/issues/827 for the merge queue decision. Thank you!

mitch-mitchel (approved):
LGTM

PR #829, opened 2026-08-25
VariableAnswer service and selector layer 

nicolearagao:
## What changed

Builds the data layer for storing answers to glossary variables, following the reads-are-selectors / writes-are-services split.

- **`variable_answer_set` service** — upserts an answer for (identity, variable), validating the value against the variable's `data_type` and `choices` first. An invalid value writes nothing.
- **Read selectors** — `variable_answer_list(identity)` and `variable_answer_map(identity, names)` in `selectors/topic_flow.py`.
- **`variable_value_validate`** — one pure, DB-free function that decides whether a value is legal. The corpus loader's `_variable_value_problem` now delegates to it, so corpus loading and answer writing can't drift apart.

## Decisions recorded

- **Reviewed is conservative.** Machine writes save `reviewed=False`; only an explicit confirmation (the guided fact page) passes `reviewed=True`. Every unconfirmed write resets the flag, even a rewrite of the same value: dropping a confirmation only re-asks a question, while a stale one would prefill a court form unchecked.
- **`in_schema` filtering implemented.** Both read selectors exclude answers to variables the corpus no longer names. Sync keeps those rows so a migration can move them, but no surface should show them. The write path is untouched, since a guard there would block that migration.
- **Cleared answers stay out of the map.** `variable_answer_map` omits `value=None` rows: prefill must see an absent key, not fill a blank with None. `variable_answer_list` keeps them for the fact page.
- **Open question for review:** `selectors/corpus.py` now imports the validator from `services/topic_flow.py`, which runs against the reads-are-selectors grain. The function only computes, never writes, but where the shared validator should live is worth settling (connects to the question of whether corpus loading belongs under `selectors/` at all).

Closes # 821

## Test plan

---

### DIFF
--- a/litigant_portal/app/selectors/corpus.py
+++ b/litigant_portal/app/selectors/corpus.py
@@ -1,12 +1,12 @@
 from __future__ import annotations
 
 import re
-from datetime import date, datetime
 from pathlib import Path
 from typing import Annotated, Literal
 
 import yaml
 from django.conf import settings
+from django.core.exceptions import ValidationError as DjangoValidationError
 from pydantic import (
     BaseModel,
     ConfigDict,
@@ -22,6 +22,7 @@ from litigant_portal.app.models.choices import (
     TopicFlowFormConditionOperator,
     VariableDataType,
 )
+from litigant_portal.app.services.topic_flow import variable_value_validate
 
 CORPUS_DIR = settings.BASE_DIR / "corpus"
 FORMS_DIR = CORPUS_DIR / "forms"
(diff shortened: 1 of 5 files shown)


nadahlberg (approved):
LGTM.

Agree that the services > selectors import is a bit of a pattern violation, but IMO the right solution would be to move corpus selectors to corpus services. I think selectors should probably be scoped to "reads from the DB" rather than "any utility that doesn't write", which would justify validation utilities living in services more generally. But approving on the grounds that refactoring the corpus code seems out of scope for this PR -- unless you want to tack it on.

nadahlberg:
@nicolearagao not for this PR, but what do you think about corpus services + selectors becoming:

```
services/corpus/schemas.py
services/corpus/load.py
services/corpus/sync.py
```

nicolearagao:
> @nicolearagao not for this PR, but what do you think about corpus services + selectors becoming:
> 
> ```
> services/corpus/schemas.py
> services/corpus/load.py
> services/corpus/sync.py
> ```



> LGTM.
> 
> Agree that the services > selectors import is a bit of a pattern violation, but IMO the right solution would be to move corpus selectors to corpus services. I think selectors should probably be scoped to "reads from the DB" rather than "any utility that doesn't write", which would justify validation utilities living in services more generally. But approving on the grounds that refactoring the corpus code seems out of scope for this PR -- unless you want to tack it on.

 Let's merge this and open a separate issue so the decision is recorded either way. 
Fair point, and I agree. If selectors are strictly "reads from the DB", then variable_value_validate belongs in services. Not a violation. I'll drop that open question from the PR description.

One thing I'd still push back on. That works for the validator, but selectors/corpus.py is ~560 lines of schemas, and those aren't business logic either. CLAUDE.md's "Open contracts" section uses the corpus schema as its example of a boundary contract. And we already have a home for that shape: app/topic_flow/ holds schema.py / loader.py / validation.py, and CLAUDE.md calls it "the corpus engine, not a data-layer module".

So why don't we go:

- app/corpus/{schema,loader}.py for the contract and parsing
services/corpus.py keeps corpus_sync, since that's the part that writes

Small loose end either way: CLAUDE.md's domain table doesn't list corpus. #814 added it without updating the table.

Issue #830, opened 2026-08-26
Where to put schemas, loaders, and validators


nadahlberg:
Following up on discussion [here](https://github.com/freelawproject/litigant-portal/pull/829#issuecomment-5427753920), something for us to chat about Friday @nicolearagao @mitch-mitchel:

## Brief recap of the discussion on PR 829

- #813 put validation logic, pydantic models, and file loading helpers for the corpus in `corpus/selectors.py`
- #829 added a shared variable value validation method to `services/topic_flow.py`, which due to the corpus validation placement, forced a services > selectors import.
- Selectors are probably more appropriately scoped as a place for "read only DB queries" rather than merely any "read only helper"
- The old topic flow engine puts validation, schemas, and loaders in a `topic_flow` module at the root of our app.

## What should we do about it?

Should we mirror the `topic_flow` module with a `corpus` module? I would push back on a few counts:

1. The `topic_flow` module pre-dates the other patterns established in this repo, and I don't love how it inverts the `{function}/{domain}` namespace pattern we use everywhere else.
2. IMO the `topic_flow` module will likely dissolve as we refactor the topic flow page to be derived from model data. Most of the submodules include methods that *do* belong in the topic flow services and selectors. The loaders, validators, schemas, and session-based answer store will go away once the topic flow page pulls from models, as these are based on the old hard-coded schemas in `content` now replaced by `corpus`. All that leaves is the renderer -- which I think we should discuss whether we want to keep given that its asymmetric with how we render other templates and I'm not sure why.

My vote would be to add new app-level modules for loaders, validators, and schemas (maybe validators and schemas go together?) and adopt the same domain-based namespacing. I.e. something like:

```
loaders/corpus.py
validators/corpus.py
... and so on ...
```

Import constraint could become something like: 
validators > loaders > selectors > services >  endpoints/views
models > selectors > services > endpoints/views

Something for us to chat about.

nadahlberg:
actually agents runs into this problem too. The agent schemas are defined in `agents/schemas` but these are imported into models for the json schema validated field for `ChatMessage`. 

Most of the logic in agents has already moved to the `chat_engine` services, selectors, and views. Maybe `agents` no longer deserves to be its own package-level module.

if we moved `tools` and `agents` under `app`, a clean path for everything could be:

validators > loaders > models > selectors > services > tools > agents > endpoints/views

PR #832, opened 2026-08-26
Assistant can activate a guided topic flow

### LINKED ISSUE #831: Assistant can activate a guided topic flow

The assistant and the guided topic flows are disconnected. The corpus pipeline syncs flows into the database and the flow pages render them, but the chat assistant has no awareness that they exist: it cannot tell a user "there is a step-by-step guide for exactly your situation," cannot ground its answers in the flow's court-specific content (deadlines, forms, local process), and has no per-thread notion of which flow a conversation is about for other surfaces to build on.

Give the assistant an active topic flow, set by the agent itself.

- LoadTopicFlow tool: takes a topic-slug/flow-slug path, validates it against enabled flows, stores it as active_topic_flow in thread state, and returns the flow's full content graph as markdown (sections, deadlines, form packet, interview variables, links) so the model grounds later answers in it. Unknown or disabled paths return an error listing the valid paths.
- System prompt lists the available flows and instructs the assistant to load one as soon as the user's situation matches. The prompt deliberately omits the active flow: it varies only with the shared enabled-flow list, never per-thread state, so all threads reuse one cached prompt artifact. The active flow reaches the model through the tool result in history instead.
- Prompt assembly moves to a PROMPT_TEMPLATE with one injection slot per section and a builder function per section, so the prompt can grow without generate_system_prompt accreting logic.
- New engine lifecycle hook, Agent.prepare_thread(thread_id): called once per user message before the message is stored or the first model call. The assistant uses it to clear a stale active flow (state stores a slug path, not a foreign key, so a synced corpus can rename, disable, or delete the flow out from under it). Documented in AGENT_DEV_GUIDE.
- Data layer: topic_flow_list and topic_flow_find selectors (enabled flows, content graph prefetched).

Known follow-up, deliberately out of scope: the agents-package import constraint that forces function-body imports (#830).

_No response_

nadahlberg:
## What changed

Gives the assistant an active topic flow, set by the agent itself. Closes #831.

- **LoadTopicFlow tool** — takes a topic-slug/flow-slug path, validates it against enabled flows, stores it as `active_topic_flow` in thread state, and returns the flow's full content graph as markdown (sections, deadlines, form packet, interview variables, links). Unknown or disabled paths return an error listing the valid paths.
- **System prompt lists the available flows** via a `PROMPT_TEMPLATE` with one injection slot and one builder function per section, so the prompt can grow without `generate_system_prompt` accreting logic.
- **`Agent.prepare_thread(thread_id)` engine hook** — called once per user message, after the thread is resolved and before the message is stored or the first model call. Default no-op; documented in AGENT_DEV_GUIDE. The assistant uses it to clear a stale active flow.
- **Selectors** — `topic_flow_list` and `topic_flow_find` (enabled flows, content graph prefetched).
- **CLAUDE.md** — adds the `corpus` domain to the data-layer table (tack-on).

## Decisions recorded

- **The system prompt omits the active flow.** It varies only with the shared enabled-flow list, never per-thread state, so all threads reuse one cached prompt artifact; a test pins the prompt as byte-identical with and without an active flow. The active flow reaches the model through the tool result in history, and the tool does not set `refresh_system_prompt`.
- **State stores a slug path, not a foreign key**, so a corpus sync can rename, disable, or delete the flow out from under it. `prepare_thread` clears a stale path on every user message; validation reuses the flow list, costing no extra query when no flow is active.
- **State writes are dict merges**, preserving unrelated keys.
- **Strict path parsing.** A pasted listing line errors with the valid paths listed, and the model self-corrects on retry; the argument description and prompt both spell out the path-only format.
-  **Model imports are nested**, a temporary bandaid to the circular imports that would result from `ChatThead` depending on `MessageSchema`. We can fix this once we find a better place for Pydantic schemas (see #830)

## Test plan

- [x] 19 new tests (tool, selectors, markdown, prompt, hook semantics, engine-loop hook ordering); fast-suite portion green, `postgres`-marked portion needs `make test`
- [x] ruff check/format clean; rebased onto #829 with the selector conflict resolved additively

---

### DIFF
--- a/litigant_portal/agents/assistant.py
+++ b/litigant_portal/agents/assistant.py
@@ -1,3 +1,5 @@
+from django.db import transaction
+
 from .base import Agent, AgentState
 from .tools.load_topic_flow import LoadTopicFlow, topic_flow_path
 from .tools.query_document import QueryDocument
@@ -71,13 +73,14 @@ class LitigantAssistant(Agent):
         from litigant_portal.app.models import ChatThread
         from litigant_portal.app.selectors.topic_flow import topic_flow_list
 
-        thread = ChatThread.objects.get(id=thread_id)
-        active = (thread.state or {}).get("active_topic_flow")
-        if active and active not in {
-            topic_flow_path(f) for f in topic_flow_list()
-        }:
-            thread.state = {**thread.state, "active_topic_flow": None}
-            thread.save(update_fields=["state", "updated_at"])
+        with transaction.atomic():
+            thread = ChatThread.objects.select_for_update().get(id=thread_id)
+            active = (thread.state or {}).get("active_topic_flow")
+            if active and active not in {
+                topic_flow_path(f) for f in topic_flow_list()
+            }:
+                thread.state = {**thread.state, "active_topic_flow": None}
+                thread.save(update_fields=["state", "updated_at"])
 
     def generate_system_prompt(self, *, thread_id) -> str:
         """Each prompt piece injected into PROMPT_TEMPLATE."""
--- a/litigant_portal/agents/assistant.py
+++ b/litigant_portal/agents/assistant.py
@@ -84,7 +84,14 @@ class LitigantAssistant(Agent):
         chat_thread_state_merge(thread_id=thread_id, updates=clear_if_stale)
 
     def generate_system_prompt(self, *, thread_id) -> str:
-        """Each prompt piece injected into PROMPT_TEMPLATE."""
+        """Each prompt piece injected into PROMPT_TEMPLATE.
+
+        The active topic flow is deliberately absent: the model learns it
+        from the LoadTopicFlow result in history, which only works while
+        the engine sends the full history. When we add automatic compaction,
+        we'll need to account for the active topic flow data being dropped from
+        history.
+        """
         return PROMPT_TEMPLATE.format(
             base=BASE_PROMPT,
             topic_flows=generate_topic_flows_prompt(),
--- a/litigant_portal/agents/assistant.py
+++ b/litigant_portal/agents/assistant.py
@@ -1,4 +1,5 @@
-from .base import Agent
+from .base import Agent, AgentState
+from .tools.load_topic_flow import LoadTopicFlow, topic_flow_path
 from .tools.query_document import QueryDocument
 
 BASE_PROMPT = """\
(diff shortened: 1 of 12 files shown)


nadahlberg:
Claude review addressed: Added `transactions.atomic` to address the read > edit > write issues and added `thread.refresh_from_db` before the `state` SSE is emitted in the stream.

litigant_portal/agents/tools/load_topic_flow.py
```diff
@@ -0,0 +1,146 @@
...
+            )
+            return ToolOutput(
+                result=(
+                    f"Error: no topic flow named '{self.topic_flow}'. "
+                    f"Available flows: {available}."
+                )
+            )
+
+        with transaction.atomic():
+            thread = ChatThread.objects.select_for_update().get(id=thread_id)
```
nicolearagao: This lock (and the one in prepare_thread) only works if every writer takes it. check_weather.py:23-28 still reads thread.state, changes it, and saves it with no lock. Its save() waits for the lock, but it already read the old value before waiting, so it can still overwrite a change made by a locked writer.

Suggestion, not a blocker: this merge-under-lock pattern now exists in two places, and a third (check_weather.py) forgot the lock. What about one service that does it for everyone?

```
# services/chat_engine.py
def chat_thread_state_merge(*, thread_id, updates: dict) -> dict:
    with transaction.atomic():
        thread = ChatThread.objects.select_for_update().get(id=thread_id)
        thread.state = {**(thread.state or {}), **updates}
        thread.save(update_fields=["state", "updated_at"])
        return thread.state
```

Then each tool calls it in one line, nobody can forget the lock, and it follows the repo rule that writes go through services (right now the tools write to ChatThread directly). This matters more soon: the v2 plan adds more state-writing tools, and each one would copy this pattern again. Fine as a follow-up issue if you'd rather keep this PR small.

litigant_portal/app/selectors/topic_flow.py
```diff
@@ -18,6 +18,35 @@ def topic_get(*, topic_id) -> Topic:
...
+def topic_flow_list() -> list[TopicFlow]:
+    """Enabled flows with their topics, in topic order then flow order."""
+    return list(
+        TopicFlow.objects.filter(enabled=True)
+        .select_related("topic")
+        .order_by("topic__order", "topic__created_at", "order", "created_at")
+    )
+
+
+def topic_flow_find(*, topic_slug: str, flow_slug: str) -> TopicFlow | None:
```
nicolearagao, line 30: Nice distinction. Can we note it somewhere (or in the docstring) that _get raises and _find returns None, so the next selector picks the right verb on purpose?

litigant_portal/agents/assistant.py
```diff
@@ -12,11 +15,76 @@
...
+        with transaction.atomic():
+            thread = ChatThread.objects.select_for_update().get(id=thread_id)
+            active = (thread.state or {}).get("active_topic_flow")
+            if active and active not in {
+                topic_flow_path(f) for f in topic_flow_list()
+            }:
+                thread.state = {**thread.state, "active_topic_flow": None}
+                thread.save(update_fields=["state", "updated_at"])
 
     def generate_system_prompt(self, *, thread_id) -> str:
```
nicolearagao, line 86: The PR body explains why the active flow is not in the prompt: the model gets it from the tool result in history. That works because we always send the full history (_messages_for_llm never trims it). But that condition only lives in the PR description. If someone later adds history trimming or summarization, the active flow silently drops out of the model's context and no test fails.

Could you add one sentence here saying the design depends on full history always being sent?

litigant_portal/app/tests/test_assistant_topic_flow.py
```diff
@@ -0,0 +1,258 @@
...
+
+    def test_malformed_paths_return_none(self):
+        _eviction_flow()
+        for path in ("eviction", "eviction/", "/tenant", ""):
+            self.assertIsNone(topic_flow_from_path(path))
+
+
+@pytest.mark.postgres
+class TopicFlowMarkdownTests(TestCase):
+    def test_markdown_covers_the_flow_content_graph(self):
```
nicolearagao, line 103: This covers the content graph well, but two branches of topic_flow_markdown never render in any test:

The asked_when line: no test variable has a gate, so (asked when X = ...) is never produced.
A form condition with a variable: the only form here is always included, so the (included when ... equals ...) line is never produced.

nicolearagao (changes requested):
Looks really good, only must from comments is test coverage.

litigant_portal/app/selectors/topic_flow.py
nadahlberg: Added a line to claude.md

litigant_portal/agents/assistant.py
nadahlberg: Added a note to the docstring of generate system prompt. We plan to add automatic compaction, where a flagged summary message will be injected into a thread, after which the history is loaded only from that point. So we'll run into the scenario you're describing and need to deal with it.

litigant_portal/agents/tools/load_topic_flow.py
nadahlberg: Added. One tweak: when we check for a stale active_topic_flow, reading the state before calling this would reintroduce the race condition transaction.atomic was trying to protect against. To address I allowed `updates` to be a dict or callable, so you can pass the reading logic to `chat_thread_state_merge` so it happens under the same transaction.

nicolearagao: accepting a dict or a callable is a great idea!!

nicolearagao (approved):
nice!

Issue #821, opened 2026-08-24
VariableAnswer service and selector layer


nicolearagao:
### What needs to change?

Build the data layer for storing answers to glossary variables, following the repo's reads-are-selectors / writes-are-services split and `{model}_{action}` naming:

- `variable_answer_set` service: upsert an answer for (identity, variable), validating the value against the variable's `data_type` and `choices` before saving. Carries the `reviewed` semantics: AI-written answers save with `reviewed=False`; a human confirming on the guided fact page flips it to `True`.
- Read selectors for the surfaces that need them (answers for an identity, answers for a set of variable names).
- Value validation must have one source of truth. The corpus schema already has `_variable_value_problem` in `selectors/corpus.py` (legal-value rules per data type); either generalize it so both the corpus loader and this service call it, or make this service's check canonical and have the corpus call it. Two slightly different copies of "is this value legal" is the failure mode to avoid.
- Decide and implement `in_schema` filtering: list selectors should exclude variables the corpus no longer names (`in_schema=False`), pending confirmation of that design on the corpus sync PR.
- Tests for all of it: validation per data type, reviewed transitions, upsert identity, in_schema filtering.

### Why?

Both pathways (chat and guided flow) store facts as VariableAnswer rows, and the reviewed flag is the trust boundary: only human-confirmed answers may enter the docassemble prefill payload, because prefilled variables skip their questions and print on court forms with no further human check. This layer is where those rules get enforced once, for every caller.

### Definition of done (optional)

- `services/` and `selectors/` modules with tests, following repo naming conventions.
- Value validation single-sourced between corpus loading and answer writing.
- reviewed=False on machine writes, reviewed=True only via explicit confirmation.
- in_schema filtering implemented or explicitly deferred with a decision recorded.

PR #833, opened 2026-08-27
feat: store guided flow answers in the database

### LINKED ISSUE #822: Guided flow answers move from session AnswerStore to VariableAnswer
### What needs to change?

Swap the guided fact page's answer storage from the session-backed `AnswerStore` (`app/topic_flow/answer_store.py`) to `VariableAnswer` rows, written through the variable_answer service:

- The `topic_flow()` view in `views/pages.py` reads and writes answers via the service layer instead of the session.
- Saving on the fact page marks answers `reviewed=True`: the guided page is the human review surface.
- The renderer keeps taking a plain dict, so it does not change.
- Page rendering itself stays on the YAML corpus engine until the Sep 3 QA build is cut; this issue changes where answers live, not where pages come from.
- Open question to settle here: guest users. AnswerStore was guest-first (session only, no DB). Decide whether guests keep a session fallback or get an identity row, and document the choice.
- Tests through real page renders, per repo convention.

### Why?

Answers stored in the session die with the browser and are invisible to the chat pathway. Storing them as VariableAnswer rows makes facts survive across visits and across pathways: a fact given in chat can appear pre-filled on the guided page, and confirmed guided answers become eligible for the docassemble prefill.

- Fact page reads and writes VariableAnswer via the service; session store retired or kept only as the guest fallback, with the choice documented.
- Answers saved on the page are reviewed=True.
- Page rendering source unchanged (still YAML until the Sep 3 build).
- Render-based tests cover read, write, and re-render with stored answers.

nicolearagao:
## What changed

Moves Topic Flow's fact_gather answers off the session-backed `AnswerStore` and onto the `VariableAnswer` table, so answers persist across sessions/devices and merge into an account on login, matching the architecture in #821/#829.

- **Question ids renamed to glossary names** — `publication_date` → `name_change_publication_date`, etc., so a fact_gather question id is the same string as its `Variable.name` everywhere (corpus YAML, renderer, validation, `_NEVER_PREFILL`). A new `test_content_glossary_agreement.py` pins that every fact_gather question id in the corpus has a matching glossary variable.
- **`views/utils.py:topic_flow_answers`** — reads `{question_id: value}` through `variable_answer_map`, for both the page render and downloads (ICS/VCF). Returns `{}` for a visitor with no session yet, so an anonymous crawler hitting a flow page doesn't mint an identity row.
- **`services/topic_flow.py:variable_answer_set_many`** — upserts a `{name: value}` batch in one transaction; an unknown name raises `Variable.DoesNotExist` (a corpus-sync failure, not litigant input) and rolls the whole batch back. The Topic Flow POST handler saves with `reviewed=True` — this page is where a human confirms what the assistant may have guessed, and only reviewed answers are eligible for the docassemble prefill.
- **`services/user.py:user_identity_merge`** — now also migrates variable answers from the anonymous identity into the target on login. `(identity, variable)` is unique, so where both identities have answered the same variable, the target's (the account's) answer wins; the anonymous one is dropped rather than overwriting a confirmed value.
- **`topic_flow/validation.py`** — adds a date-format check (`YYYY-MM-DD`) at the same soft-gate layer as the existing required/choice checks, so an unparseable date surfaces as an inline "fix this" instead of a 500 at the storage layer, which now persists dates typed.
- Removed `topic_flow/answer_store.py` (session-backed `AnswerStore`) — fully replaced.

Closes #822 

## Test plan

---


nadahlberg (approved):
Looks good to me. One issue is that NEVER_PREFILLS prevents `name_change_publication_date` from prepopulating the form. So if a user submits the publication date, then submits the form again without filling out the publication date, we will lose the data. 

I'm inclined to merge this one anyway and open a separate issue for us to decide what data we actually want to retain and use for anon users. NEVER_PREFILLS feels like a one-off hack (the calculated offset date shows anyways at a different part of the page) and doesn't really answer the question about how we want to handle session hygiene for anon users.

nicolearagao:
> Looks good to me. One issue is that NEVER_PREFILLS prevents `name_change_publication_date` from prepopulating the form. So if a user submits the publication date, then submits the form again without filling out the publication date, we will lose the data.
> 
> I'm inclined to merge this one anyway and open a separate issue for us to decide what data we actually want to retain and use for anon users. NEVER_PREFILLS feels like a one-off hack (the calculated offset date shows anyways at a different part of the page) and doesn't really answer the question about how we want to handle session hygiene for anon users.

Agreed, let's merge. I confirmed the loss: the field is optional, so a blank submit clears the answer, and _NEVER_PREFILL guarantees the form is blank on re-render. Note the guard doesn't hide much anyway, since the 30-day deadline on the page lets anyone work the date back out. Filed #835 to decide anon answer retention properly.

PR #834, opened 2026-08-27
feat(ai): render prompt artifacts in Markdown audit transcripts

### LINKED ISSUE #840: Render prompt artifacts in Markdown audit transcripts
### What needs to change?

The Markdown-transcript slice of #790: render captured prompt artifacts in the Markdown audit download, so a reviewer can read the instruction state without cross-referencing the JSON export.

- A `Prompt artifact` block rendered immediately before its first referenced assistant message, and again whenever the active non-null `prompt_artifact_id` changes.
- Last-known artifact state preserved across user, tool, hidden, meta, and legacy rows carrying null references, so a null does not reset the renderer.
- Transitions back to an earlier artifact (A → B → A) render again rather than being suppressed.
- Prompt-artifact transitions kept independent of deployed-SHA transitions; when both change on one turn, the SHA change renders first, then the artifact, then the assistant message.
- Collision-safe Markdown fences, so a system prompt containing headings or its own code fences cannot break the transcript structure.
- Tool schemas pretty-printed for readable audit review.
- `docs/wiki/audit-transcripts.md` updated to describe Markdown prompt artifacts, null-reference behavior, and the limits that remain.

### Why?

#825 delivered the storage and JSON-export half and explicitly left Markdown transcript rendering to #790. Reading an instruction state currently means opening the JSON download and resolving IDs by hand, which is the wrong ergonomics for the audit surface staff actually use.

Splitting this out gives the work in flight an issue it can close, and leaves #790 open for the rest rather than closing it early.

- [ ] Markdown transcript renders a prompt artifact on first use and on every change of the active non-null reference
- [ ] Null references do not reset the renderer's last-known artifact state
- [ ] A → B → A renders the returning artifact
- [ ] Deployed-SHA and prompt-artifact transitions stay independent, SHA first when both change
- [ ] A system prompt containing code fences cannot break transcript structure
- [ ] Audit-transcript docs describe the Markdown behavior and the remaining limits

Explicitly out of scope, still tracked on #790: tool-internal prompt capture, orphan cleanup and retention for unreferenced artifacts, model-name tracking.

zeitcow:
## What changed

- Render captured prompt artifacts in Markdown audit transcripts.
- Show the full instruction state, including the rendered system prompt and tool-schema snapshot.
- Render a prompt artifact immediately before its first referenced assistant message and again whenever the active non-null artifact changes.
- Preserve the last-known artifact state across user, tool, hidden, meta, and legacy rows with null prompt references.
- Correctly render transitions back to an earlier artifact, such as A → B → A.
- Keep prompt-artifact transitions independent from deployed-SHA transitions.
- When both change on the same turn, render the deployed-SHA change first, then the prompt artifact, then the assistant message.
- Use collision-safe Markdown fences so prompts containing headings or code fences cannot break the transcript structure.
- Pretty-print tool schemas for readable audit review.
- Update the audit-transcript documentation to describe Markdown prompt artifacts, null-reference behavior, and the remaining retention/tool-internal-prompt limits.
- Add focused coverage for first use, unchanged artifacts, changed artifacts, returning artifacts, null rows, legacy messages, tool-call-only assistant responses, schema-only changes, and Markdown/SHA ordering.

This is the Markdown transcript slice of #790.

It does not change prompt capture, storage, hashing, deduplication, JSON export behavior, models, migrations, tool-internal prompt capture, orphan cleanup/retention, model-name tracking, or deployed-SHA capture.

Part of #790

## Test plan

Validated against the PostgreSQL-backed Docker environment:

- [x] Audit-export test module: 28 passed
- [x] Ruff check passed for changed Python files
- [x] Ruff format check passed for changed Python files
- [x] Prettier passed for `docs/wiki/audit-transcripts.md`
- [x] `git diff --check`

## AI Disclosure

- [ ] No AI tools were used to create the content of this PR.
- [x] Parts of this PR were created with the help of an AI tool, and I have carefully reviewed all of its content and take full responsibility for it.

AI model used:

- ChatGPT GPT-5.6 Sol via Codex

I reviewed the implementation, tests, Markdown rendering behavior, null-reference state handling, documentation, Docker test results, and final diff, and take responsibility for the contribution.

Closes #840

### DIFF
--- a/docs/wiki/audit-transcripts.md
+++ b/docs/wiki/audit-transcripts.md
@@ -15,7 +15,7 @@ You need a portal account with **staff status**. A developer grants this once (i
    - **Filter** by date or thread type using the sidebar.
 4. Click the thread to open it. The page shows who the conversation belongs to, when it started, and the full transcript: user messages, AI answers, tool calls the AI made, and tool results. Rows marked `[hidden]` or `[meta]` were not visible to the user; they are included because an audit needs the complete record.
 5. To save a copy, use the **Downloads** links on the same page:
-   - **Markdown**: a readable transcript, good for review and sharing.
+   - **Markdown**: a readable transcript, including captured instruction states on first use and whenever they change, good for review and sharing.
    - **JSON**: the raw record with timestamps, token counts, and cost, good for deeper analysis or archiving.
 
 Everything is read-only. You cannot change or delete a conversation from this screen.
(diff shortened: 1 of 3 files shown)


mitch-mitchel:
Read the full diff — this is clean, and the care shows in the details.

`_fenced_block` gets the fence math right: longer than any internal backtick run with a floor of 3, so a prompt containing its own code fences cannot break the transcript. The A → B → A case re-renders because `active_prompt_artifact_id` tracks the last *rendered* artifact rather than a seen-set, and null references genuinely do not reset it. Ordering matches the description: SHA change, then artifact, then message.

I also checked whether `artifacts_by_id[prompt_artifact_id]` could KeyError on a dangling reference — it cannot. Both the lookup and the message's `prompt_artifact_id` derive from the same message list in `chat_thread_export_data`, and the FK is `on_delete=PROTECT`.

The doc update is the part I appreciate most: it adds the "unreferenced prompt artifacts are not collected yet" limit rather than leaving it implicit, and it is accurate about null-reference behavior.

Two housekeeping notes, neither on the code:

1. This had no closing keyword because it is a slice of #790, which is the right call — but our convention is one issue per PR, and without a link the board automation cannot move anything. I have filed #840 for exactly this slice (mirroring how #825 was split out for the storage half). **Could you add `Closes #840` to the PR description?** I would rather not edit your PR body myself.
2. Base is 6 commits behind main; I will update the branch before merging.

Thanks for the thorough test plan — the enumerated cases are exactly the ones that matter here.

zeitcow:
Thanks, Mitch. I added `Closes #840` to the PR description. I appreciate the thorough review.

nicolearagao (approved):
Reviewed this as well, and I have nothing to add that would change the code. To avoid repeating Mitch's review, I checked the parts it didn't mention:

- The unconditional msg["prompt_artifact_id"] read is safe for every row type: chat_thread_export_data sets the key on all message rows (explicitly None when unset), and both the lookup keys and the references are str(uuid) from the same list, so no type mismatch either.
- Tool schemas go through the same collision-safe fence as the system prompt, so a backtick-heavy tool description can't break the transcript.
- The test matrix covers the two cases that are easiest to forget here: a tool-call-only assistant message, and a schema-only artifact change. Nice.
Thanks for the careful work on this, the coverage and the doc update are exactly what the audit story needed.

mitch-mitchel (approved):
Looks good!, :shipit:

PR #837, opened 2026-08-28
Consolidate model choices around AWS Bedrock

### LINKED ISSUE #685: Remove dead CHAT_ENABLED / DEFAULT_CHAT_AGENT / CHAT_MODEL config plumbing
`settings.py:348-355` defines `CHAT_ENABLED`, `DEFAULT_CHAT_AGENT`, and `CHAT_MODEL`, but their only consumers were chat/agents v1, deleted in #682. Model selection is DB-driven now (`site_get_model()` / `get_default_model()`). Setting these vars silently does nothing, and `DEFAULT_CHAT_AGENT`'s default names a class that no longer exists.

Still plumbed through: `docker-compose.yml:38-40`, `deploy/qa-do/docker-compose.qa.yml:81-82` (whose header instructs operators that the QA `.env` **must** set `CHAT_MODEL`), `tox.ini` (`CHAT_ENABLED=false` in both testenvs), and `.env.example:27-28`.

**Fix:** delete from all five places. Keep `OPENAI_API_KEY` (LiteLLM + `get_default_model()` read it). If a chat kill switch is still wanted, reintroduce `CHAT_ENABLED` only when something reads it.

### LINKED ISSUE #836: Consolidate model choices around AWS Bedrock
### What needs to change?

Consolidate the app's model configuration on AWS Bedrock as the single provider:

- Delete the `OpenAIModel` enum and the grouped `AI_MODEL_CHOICES`; the `Site` model's `fast_model` / `assistant_model` fields take `BedrockModel.choices` directly (choices-only migration, no db operation).
- Replace the provider-sniffing `get_default_model()` with a `DEFAULT_BEDROCK_MODEL` constant (GPT-5.6 Luna).
- Admin settings UI: drop the dynamic provider logic; always show the flat Bedrock model list regardless of credentials, keep the "Bedrock available" badge driven by `AWS_BEARER_TOKEN_BEDROCK`, and strip all OpenAI references.
- Uploads: remove the OpenAI text-extraction fallback in attachment hydration; all document types ship as native Bedrock document blocks. This also removes the now-dead per-request text budget and the `model` parameter threading, and fixes the stale `bedrock/` prefix check that no longer matched the `bedrock_mantle/` model ids (which was silently routing every document through the extraction path).
- Env/docs sweep: `AWS_BEARER_TOKEN_BEDROCK` becomes the required credential in `.env.example`, compose files, READMEs, and CLAUDE.md; all `OPENAI_API_KEY` references removed.
- Includes the #685 cleanup: the dead `CHAT_ENABLED` / `DEFAULT_CHAT_AGENT` / `CHAT_MODEL` settings and their plumbing (compose files, tox.ini, .env.example, docs) are deleted rather than repointed at Bedrock. Note #685 suggested keeping `OPENAI_API_KEY`; that no longer applies since `get_default_model()` is gone and nothing reads the key.

### Why?

All models are now served through Bedrock, so the two-provider machinery (grouped choices, credential-dependent defaults, per-provider upload paths) is dead weight, and part of it was actively broken: the upload path's `bedrock/` prefix check predates the `bedrock_mantle/` ids, so native document blocks were unreachable. One provider means one hydration path, one default, and a simpler admin surface.

- No `OpenAIModel`, `AI_MODEL_CHOICES`, `get_default_model`, or `OPENAI_API_KEY` references outside `archive/`.
- Admin settings always lists Bedrock models; badge reflects credential presence only.
- Office/text/pdf attachments ship as native document blocks; rtf is the only stubbed-as-unreadable type.
- #685's five plumbing sites are clean (closes #685).

nadahlberg:
Fixes #836. Closes #685.

All models are served through Bedrock now, so this removes the two-provider machinery end to end:

## Model layer
- `OpenAIModel` and the grouped `AI_MODEL_CHOICES` are gone; `Site.fast_model` / `Site.assistant_model` take `BedrockModel.choices` directly (choices-only migration, no db operation).
- `get_default_model()`'s credential sniffing is replaced by `DEFAULT_BEDROCK_MODEL` (GPT-5.6 Luna).
- litellm bumped 1.88.0 → 1.98.0 for `bedrock_mantle` model ids.

## Admin settings
- The model selects always list the flat Bedrock choices, selectable regardless of credentials; the green "Bedrock available" badge still reflects whether `AWS_BEARER_TOKEN_BEDROCK` is set. All OpenAI references stripped.

## Uploads
- Single hydration path: every document type ships as a native Bedrock document block; only images remain image parts. The OpenAI text-extraction fallback, the per-request text budget, and the `model` parameter threading are deleted (text files now spend doc slots).
- Fixes a live bug: the `bedrock/` prefix check predated the `bedrock_mantle/` ids, so every office/text file was silently taking the extraction path and native document blocks were unreachable.
- **Behavior note:** small `.txt`/`.csv`/`.md` attachments previously inlined as plain text in the prompt; they now ride as document blocks. Worth a smoke test with a text attachment.

## Env & docs (#685)
- Dead `CHAT_ENABLED` / `DEFAULT_CHAT_AGENT` / `CHAT_MODEL` deleted from settings, both compose files, tox.ini, and `.env.example` (their consumers went with chat v1). #685 said to keep `OPENAI_API_KEY`, but that no longer applies: `get_default_model()` is gone and nothing reads the key.
- `AWS_BEARER_TOKEN_BEDROCK` is the required credential everywhere (`.env.example`, READMEs, CLAUDE.md); local compose passes it through instead of the AWS key triple.

## Deploy note
The QA box's `/opt/litigant-portal/.env` needs `AWS_BEARER_TOKEN_BEDROCK` added before the next deploy; its `CHAT_MODEL`/`OPENAI_API_KEY` entries become harmless no-ops.


litigant_portal/app/migrations/0018_bedrock_model_choices.py
```diff
@@ -0,0 +1,48 @@
...
+                        "bedrock_mantle/anthropic.claude-haiku-4-5",
+                        "Claude Haiku 4.5",
+                    ),
+                    ("bedrock_mantle/zai.glm-4.7-flash", "GLM 4.7 Flash"),
+                ],
+                max_length=128,
+                null=True,
+            ),
+        ),
+    ]
```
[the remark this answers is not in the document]
nadahlberg: Updated the migration for backwards compat

mitch-mitchel:
Reviewed the full diff. The migration is the strongest part: `_clear_stale_models` catches that `AlterField(choices=...)` never touches stored data, so a site sitting on `openai/gpt-5-nano` would survive the migration and then fail at litellm, because `site_get_model` falls back only on empty values, not stale ones. Correct ordering, `elidable=True`, noop reverse. Nice catch on the `bedrock/` vs `bedrock_mantle/` prefix bug too — native document blocks were unreachable.

Verified separately: PDFs still ship native (`PDF_TYPE` is in `BEDROCK_DOC_TYPES`), and there is no migration collision (main's leaf is `0017`, this adds `0018`, no other open PR carries one).

**One thing to confirm before merge.** `docker-compose.yml` drops the `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN` triple for `AWS_BEARER_TOKEN_BEDROCK`. The deploy note covers the QA box, but prod auto-deploys on push to main and sources its runtime env from `litigant-env`. Does that resource already carry `AWS_BEARER_TOKEN_BEDROCK`? If it is still on the key triple, chat breaks in prod on merge.

**Two questions, neither blocking:**

1. `get_default_model()` deliberately returned the smallest model; `DEFAULT_BEDROCK_MODEL` is now Luna, and `site_get_model` uses that one default for both roles — so `fast_model` defaults to a frontier model where it used to default to Haiku 4.5. Intentional? Haiku 4.5 is still in the roster.
2. `REQUEST_DOC_BUDGET` stays at 4 while text types now compete for those slots (they had their own 120k-char budget before). `test_text_files_consume_document_slots` pins the behavior honestly — is 4 still the right number now that it absorbs text?

Also: base is 6 commits behind main, so worth an update-branch before merge. And #836 has no size, priority, or milestone.

mitch-mitchel:
One more thing worth calling out, spotted from a running container rather than the diff: this PR carries a **Django minor version bump** that the description does not mention.

- `main` `uv.lock` pins Django **6.0.6**
- this branch's `uv.lock` pins Django **6.1**

The body notes the litellm 1.88 → 1.98 bump for the `bedrock_mantle` ids, so the Django jump is presumably fallout from re-resolving the lock rather than an intentional upgrade. Either way it is a framework minor riding inside a provider-consolidation PR, and `CLAUDE.md` still says Django 6.0 — worth a line in the description so a reviewer is told rather than surprised.

Two consequences, both fine, just worth being deliberate about:

1. **It resolves our three open Dependabot advisories.** All three are Django runtime alerts (2 medium, 1 low) patched in 6.0.7, so 6.1 clears them. They have been open 21 days because #783 (the 6.0.6 → 6.0.7 bump) was closed back in August to force a rescan, which Dependabot read as "ignore this release." If this lands, #783 becomes moot; if this sits, we should reopen #783 rather than wait.
2. **Nothing in CI exercises 6.1 against a real Bedrock call**, since tests mock all providers. Combined with the `litigant-env` credential question above, the first real exercise of both the new provider path and the new Django minor is the production deploy on merge.

No action needed from you on the advisories — flagging so the version change is a decision rather than a side effect.

mitch-mitchel:
Went through the rest of the diff line by line. **The code all matches the prose** — no hidden behavior changes. Specifically checked and cleared:

- `views/pages.py` and `views/admin.py` — the provider-group logic and `valid_models` union are gone cleanly, no leftovers
- `query_document.py` — `model` is still used (passed to `self.ask`), so dropping it from `user_upload_content_part` left no dead local
- `models/site.py`, `test_site_singleton.py`, `test_chat_hidden.py` — consistent, and the singleton test's comment correctly re-justifies why `GPT_5_6_SOL` (not the default) is the right value to assert against
- `deploy/qa-do/*`, `README.md`, `UPLOAD_SYSTEM.md` — accurate, including dropping the 120k-char text budget from the documented budget list
- PDFs still ship native (`PDF_TYPE` is in `BEDROCK_DOC_TYPES`), and there is no migration collision (main's leaf is `0017`, this adds `0018`, no other open PR carries one)

**The gap is entirely `uv.lock`.** It is not a targeted litellm bump — it is a full re-resolution: **73 packages change version, 9 of them major, 5 added, 8 removed.** The description mentions one. At +1183/−940 the individual lines are effectively unreviewable, which is why this is worth naming rather than absorbing.

The ones with real consequences:

**Production runtime**
- **`gunicorn 25.0.3 → 26.2.0` (major)** — this is the server that runs `web-prod`. Nothing in CI exercises gunicorn; CI runs pytest. First execution is the prod deploy.
- `django 6.0.6 → 6.1` (per my previous comment)
- `django-allauth 65.14.1 → 65.19.1` — login and auth
- `django-cotton 2.6.1 → 2.7.2` — the entire component layer
- `psycopg 3.3.3 → 3.3.4`, `redis 8.0.0 → 8.1.0` — DB and cache drivers
- `pydantic 2.12.5 → 2.13.5`, `pydantic-core 2.41.5 → 2.46.5`, plus a new `pydantic-settings` — `django-pydantic-field` backs `ChatMessage.data` via `SchemaField(schema=MessageSchema)`, so this sits under message serialization

**One that is domain-specific to us: `tzdata 2025.3 → 2026.3` (major).** We compute filing deadlines and emit ICS files. A timezone-database change is precisely the kind of thing that can move a computed date. Almost certainly fine, but worth running the deadline tests and eyeballing one generated `.ics` before merge.

**Dev tooling (no prod risk, but it will change what everyone sees)**
- `mypy 1.19.1 → 2.3.1` (major), `ruff 0.15.0 → 0.16.5`, `djlint 1.36.4 → 1.44.2` — a `make lint` run after this lands may report or reformat differently than before
- `tox 4.34.1 → 4.60.1`, `tox-uv 1.29.0 → 1.36.0` plus a new `tox-uv-bare`, with `typer`/`rich`/`shellingham` dropped — every dev's existing `.tox` env goes stale and needs `tox -e fast --recreate`
- Remaining majors: `attrs`, `cssbeautifier`/`jsbeautifier` 1.15.4 → 2.0.3, `rpds-py` 0.30.0 → 2026.6.3, `virtualenv` 20.36 → 21.7, `zipp` 3.23 → 4.1

**The ask:** could the lock upgrade come out of this PR? Either `uv lock --upgrade-package litellm` (plus whatever else `bedrock_mantle` genuinely requires), leaving the rest to Dependabot, or land the wholesale upgrade as its own PR.

The reason is diagnostic, not procedural. As it stands, if prod misbehaves after merge, we cannot tell whether it was the provider consolidation or one of 73 dependency bumps — and a gunicorn major plus a Django minor on the deploy path is exactly where that ambiguity costs most. Separated, each half stays revertible on its own.

litigant_portal/app/services/upload.py
```diff
@@ -273,35 +268,21 @@ def _file_part(upload: UserUpload, data: bytes) -> dict[str, Any]:
 
 
 def user_upload_content_part(
```
nicolearagao, line 270: Ran the text-attachment smoke test the description asks for, against bedrock_mantle/openai.gpt-5.6-luna on this branch. It fails: text attachments reach the model, but the model doesn't perceive them as attachments.

Attaching a lease.txt containing secret: BANANA-42 and asking "what's the secret in my file?" gets: "I don't see an attached file in this conversation... Please upload it again." The assistant then warns the user about pasting a password in chat. It repeats across rephrasings.

Hydration is correct, so this isn't a wiring bug. The exact part built is
`{"type": "file", "file": {"file_data": "data:text/plain;base64,c2VjcmV0OiBCQU5BTkEtNDIK", "filename": "lease.txt"}}`

Probing the model directly with "Did I attach a file? If so give its exact filename":

.txt document block → "No file appears to be attached"
.csv document block → "No file was attached. The text you provided is: ..."
.pdf document block → "Yes, you attached..." (registers as an attachment)
.txt as the old labeled text part → "The secret in your file is: BANANA-42"
So the content does arrive (the model can quote it), but for text MIME types the attachment framing is lost and the model attributes the content to the user's own typing. The old path's Attached file "{name}" (upload_id=...) wrapper was doing real work here, and dropping it is what regressed.

It's also phrasing-dependent, which is the part I'd most want to avoid shipping: "what's the secret in the attached file?" works, "what's the secret in my file?" fails. 

Would it be worth keeping a short text label alongside the document block for TEXT_TYPES (name + upload_id, without re-introducing the extraction fallback or the char budget)? That would preserve this PR's single-hydration-path goal while keeping the framing the model needs. PDFs and office docs look fine as-is.

ps: I tested only default model.

nicolearagao (changes requested):

mitch-mitchel:
Heads up — #846 (pypdf 6.15.0 → 6.16.1) landed on main yesterday and conflicted this branch on `uv.lock`. I merged main in for you rather than leave you a rebase on top of the review round: `96455b4`. Full pre-commit passed before the merge commit, and the PR is back to just needing review.

What I decided, so you can override it if I got it wrong:

- **`uv.lock` takes main's copy wholesale.** The branch's lock diff was +1183/−940 with no `pyproject.toml` change, so it read as an `--upgrade` rather than something this PR needs — which also settles the undisclosed Django 6.0.6 bump I asked about earlier. If a dependency change does belong here, add it to `pyproject.toml` and re-lock; it'll show up as an intentional two-line diff instead of a thousand.
- **`CLAUDE.md`, `views/pages.py` and `test_upload.py` auto-merged** with no conflict. I checked the last one deliberately, since #845 added tests to the same file you're deleting from: all nine names it imports still resolve after your upload-service changes, and nothing references the removed `REQUEST_TEXT_BUDGET` or `_is_bedrock`.

I didn't touch anything else — no review feedback addressed, no test content edited. Still yours:

- Nicole's requested changes
- The one actually blocking: does prod's `litigant-env` carry `AWS_BEARER_TOKEN_BEDROCK`? Compose drops the AWS key triple in this PR and prod auto-deploys on push to main, so that answer gates the merge rather than the code — the bearer-token path itself is proven locally.

litigant_portal/app/services/upload.py
nicolearagao: Fixed on the branch, and re-ran the smoke tests against the default model:

The root cause was the block type, not just the missing label. Text/* document blocks fail in both paths: inline, the model attributes the content to the user's own typing, and through query_document the reader answers "the contents were not provided". So a label alone wasn't enough (tried it first — the model then acknowledged the attachment but still couldn't connect it to the content).
TEXT_TYPES now ship as a single labeled text part: [Attached file "<name>" (<type>, upload_id=...)] header plus the decoded content. No extraction fallback and no separate char budget — text decode is one line, and the existing 40k-char small-file gate already bounds it. query_document builds parts through the same function, so the reader path is fixed by the same change.
Verified: "what's the secret in my file?" (the phrasing that failed) now answers correctly for .txt, both inline and via the tool.

nicolearagao:
> Heads up — #846 (pypdf 6.15.0 → 6.16.1) landed on main yesterday and conflicted this branch on `uv.lock`. I merged main in for you rather than leave you a rebase on top of the review round: `96455b4`. Full pre-commit passed before the merge commit, and the PR is back to just needing review.
> 
> What I decided, so you can override it if I got it wrong:
> 
> * **`uv.lock` takes main's copy wholesale.** The branch's lock diff was +1183/−940 with no `pyproject.toml` change, so it read as an `--upgrade` rather than something this PR needs — which also settles the undisclosed Django 6.0.6 bump I asked about earlier. If a dependency change does belong here, add it to `pyproject.toml` and re-lock; it'll show up as an intentional two-line diff instead of a thousand.
> * **`CLAUDE.md`, `views/pages.py` and `test_upload.py` auto-merged** with no conflict. I checked the last one deliberately, since [test: add coverage for chat engine tool execution, thread endpoints, and upload serialization #845](https://github.com/freelawproject/litigant-portal/pull/845) added tests to the same file you're deleting from: all nine names it imports still resolve after your upload-service changes, and nothing references the removed `REQUEST_TEXT_BUDGET` or `_is_bedrock`.
> 
> I didn't touch anything else — no review feedback addressed, no test content edited. Still yours:
> 
> * Nicole's requested changes
> * The one actually blocking: does prod's `litigant-env` carry `AWS_BEARER_TOKEN_BEDROCK`? Compose drops the AWS key triple in this PR and prod auto-deploys on push to main, so that answer gates the merge rather than the code — the bearer-token path itself is proven locally.

Addressed the review round on the branch:

litellm re-bump: taking main's uv.lock wholesale in 96455b4 dropped the litellm upgrade this PR needs - bedrock_mantle/* ids require >= 1.98 and the lock was back at 1.88. Re-added as the targeted change you asked for: litellm>=1.98 in pyproject.toml plus uv lock --upgrade-package litellm (1.99.0, plus pydantic-settings which turns out to be litellm's own new dependency). Smoke tests below ran on 1.99, so the bedrock_mantle path is verified on it.
Fast model default: agreed it was unintentional fallout - added DEFAULT_FAST_BEDROCK_MODEL (Claude Haiku 4.5) and made site_get_model fall back per role. Configured values are untouched.
REQUEST_DOC_BUDGET = 4: proposing to keep it. Text attachments are typically tiny, and the 16 MB byte budget is the real ceiling; raising the slot count without evidence would be guesswork. Easy to revisit if real threads hit the limit.
Text attachments: the smoke test from the PR description found a real regression - see the resolved thread on upload.py for details. Short version: text/* document blocks don't work on the default model (misattributed to user typing, unreadable by the reader), so TEXT_TYPES now ship as one labeled text part; PDFs and office docs stay native blocks (both verified working).
Still open before merge: does prod's litigant-env carry AWS_BEARER_TOKEN_BEDROCK? That one needs someone with access to the prod env resource.

PR #839, opened 2026-08-28
Court context section in the assistant system prompt

### LINKED ISSUE #838: Court context section in the assistant system prompt

The assistant's system prompt says nothing about which court the portal is deployed for. The Site singleton already carries the court config (`court_name`, `jurisdiction_level`, `state`, `official_url`, `official_resources_url`), but none of it reaches the model, so answers can't be grounded in the deployment's jurisdiction. There is also no signal for the dev/demo setup where the corpus is synced with no court selected and the loaded topic flows span multiple courts.

Part of #179.

Add a `## Court context` section to `LitigantAssistant.generate_system_prompt()`, built from the Site singleton:

- When `court_name` is populated: state the court the assistant is operating in, plus jurisdiction level, state, and the official URLs, each line included only when populated.
- When `court_name` is blank (a multi-court sync leaves the site's court fields alone): render a multi-court variant instead, telling the assistant the guided topic flows may belong to different courts and to confirm the user's court and state before relying on court-specific details.

No sync or model changes; `sync_corpus --court` help text gains a sentence noting a blank court name is treated as multi-court mode.

_No response_

nadahlberg:
Fixes #838. Part of #179.

## What

Adds a `## Court context` section to the litigant assistant's system prompt, built from the Site singleton's court config:

- When `court_name` is populated, the section states the court the assistant is operating in, plus jurisdiction level, state, official court website, and official self-help resources URL. Each line is included only when its field is populated.
- When `court_name` is blank (a no-court `sync_corpus` run leaves the site's court fields alone), the section renders a multi-court variant instead: the guided topic flows may belong to different courts, so confirm the user's court and state before relying on court-specific details.

`generate_system_prompt` now joins the non-empty sections directly, which also removes the stray blank lines the old fixed template produced when a section was empty.

No sync or model changes. The `sync_corpus --court` help text gains one sentence noting that a blank court name is treated as multi-court mode.

## Tests

Four new cases in `AssistantSystemPromptTests`: fully populated court config, blank optional fields omitted, and the multi-court variant on a blank court name. All are `postgres`-marked, so they run under `make test`.

### DIFF
--- a/litigant_portal/agents/assistant.py
+++ b/litigant_portal/agents/assistant.py
@@ -13,6 +13,16 @@ files appear directly in the conversation. A note reading [Attached file \
 tool with its upload_id to read or query it. Never guess at the contents \
 of a file you haven't seen."""
 
+COURT_PROMPT = """\
+## Court context
+
+{context}"""
+
+MULTI_COURT_CONTEXT = """\
+This portal is running in multi-court mode: the guided topic flows may \
+belong to different courts. Confirm which court and state the user's case \
+is in before relying on court-specific details."""
+
 TOPIC_FLOWS_PROMPT = """\
 ## Guided topic flows
 
@@ -26,10 +36,28 @@ flow is active. Available flows:
 
 {flows}"""
 
-PROMPT_TEMPLATE = """\
-{base}
 
-{topic_flows}"""
+def generate_court_prompt() -> str:
+    """The court-context section. A blank court name means the site
+    wasn't synced to one court, so the flows may span several."""
+    from litigant_portal.app.selectors.site import site_get
+
+    site = site_get()
+    if not site.court_name:
+        return COURT_PROMPT.format(context=MULTI_COURT_CONTEXT)
+    lines = [f"You are operating in {site.court_name}."]
+    if site.jurisdiction_level:
+        level = site.get_jurisdiction_level_display()
+        lines.append(f"- Jurisdiction level: {level}")
+    if site.state:
+        lines.append(f"- State: {site.get_state_display()}")
+    if site.official_url:
+        lines.append(f"- Court website: {site.official_url}")
+    if site.official_resources_url:
+        lines.append(
+            f"- Court self-help resources: {site.official_resources_url}"
+        )
+    return COURT_PROMPT.format(context="\n".join(lines))
 
 
 def generate_topic_flows_prompt() -> str:
@@ -84,7 +112,7 @@ class LitigantAssistant(Agent):
         chat_thread_state_merge(thread_id=thread_id, updates=clear_if_stale)
 
     def generate_system_prompt(self, *, thread_id) -> str:
-        """Each prompt piece injected into PROMPT_TEMPLATE.
+        """The non-empty prompt sections, blank-line separated.
 
         The active topic flow is deliberately absent: the model learns it
         from the LoadTopicFlow result in history, which only works while
@@ -92,7 +120,12 @@ class LitigantAssistant(Agent):
         we'll need to account for the active topic flow data being dropped from
         history.
         """
-        return PROMPT_TEMPLATE.format(
-            base=BASE_PROMPT,
-            topic_flows=generate_topic_flows_prompt(),
-        ).strip()
+        return "\n\n".join(
+            section
+            for section in (
+                BASE_PROMPT,
+                generate_court_prompt(),
+                generate_topic_flows_prompt(),
+            )
+            if section
+        )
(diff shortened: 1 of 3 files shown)


nadahlberg:
I decided to make the multi-court mode part of the prompt depend on the `Site.court_name`. This way our dev deploy, which leaves `CORPUS_COURT` blank defaults to multi-court mode -- which makes sense for testing multiple topic flows from different courts (model gets confused if its doing a flow that references court specific documents while having a different court in its system prompt).

However, the reason I used `Site.court_name` instead of `CORPUS_COURT` directly, is that this gives us the option to manually set `court_name` and other court details on the admin site if we ever want to test out the effects of the court-metadata awareness. Does that make sense / seem like the right approach?

litigant_portal/agents/assistant.py
```diff
@@ -26,10 +36,28 @@
...
 
-{topic_flows}"""
+def generate_court_prompt() -> str:
+    """The court-context section. A blank court name means the site
+    wasn't synced to one court, so the flows may span several."""
+    from litigant_portal.app.selectors.site import site_get
+
+    site = site_get()
+    if not site.court_name:
+        return COURT_PROMPT.format(context=MULTI_COURT_CONTEXT)
```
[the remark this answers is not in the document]
nadahlberg: See my comment below. I think this is ok. The idea is that on the dev site we should be able to switch between multi-court mode and simulating a specific court as we see fit. For court-partner deploys, `COURT_CONFIG` will always be set. And we aren't exposing the admin site to those partners yet.

litigant_portal/app/tests/test_assistant_topic_flow.py
```diff
@@ -238,6 +239,41 @@ def test_no_flow_section_when_no_flows_exist(self):
...
         self.assertNotIn("Guided topic flows", prompt)
 
+    def test_court_context_from_site_config(self):
+        site = Site.objects.get()
+        site.court_name = "Alpha District Court"
+        site.jurisdiction_level = "state"
+        site.state = "ND"
+        site.official_url = "https://alpha.test"
+        site.official_resources_url = "https://alpha.test/help"
+        site.save()
```
nicolearagao, line 249: Could these three tests write the Site through the service instead of a raw site.save()? We already have the pattern in test_site_singleton.py:
```
with self.captureOnCommitCallbacks(execute=True):
    site_update(
        court_name="Alpha District Court",
        jurisdiction_level="state",
        state="ND",
        official_url="https://alpha.test",
        official_resources_url="https://alpha.test/help",
    )
```
Two reasons:

Test convention here is to create data through the service layer when one exists, and site_update is what the admin view calls.

site_get caches the Site row forever, and a raw ORM save never clears that cache. The tests pass today only because nothing has read the site yet when they run. If setUp (or the agent itself) ever calls site_get first, the blank site gets cached and these tests fail in a confusing way. Going through site_update inside captureOnCommitCallbacks(execute=True) clears the cache properly, so the tests can't be bitten by that.


Same applies to test_court_context_omits_blank_fields below. test_blank_court_name_means_multi_court_mode doesn't write at all, so it's fine as is.

nicolearagao (changes requested):

litigant_portal/app/tests/test_assistant_topic_flow.py
nicolearagao: pushed test that resolves this

nicolearagao (approved):

mitch-mitchel (approved):
Reviewed the full diff. Approving.

Checked and cleared:

- **Function-local `site_get` import** in `generate_court_prompt()` is the established pattern in this module (`generate_topic_flows_prompt()` and `prepare_thread()` do the same), and necessary rather than stylistic: `app` imports `agents` at module level, so a top-level import here is a cycle.
- **No prompt-level cache to stale out.** `assistant.py` holds no cache, so the prompt is rebuilt per call and `site_get()` reads live. Combined with `site_update`'s `@busts_cache(SITE_CACHE_KEY)` bust at commit, an admin court-config change reaches the next generated prompt. This was the main thing I wanted to confirm, since the prompt now depends on Site config where before it depended only on the enabled-flow list.
- **The dropped `.strip()`** is a no-op here. `BASE_PROMPT`, `COURT_PROMPT`, `MULTI_COURT_CONTEXT`, and the `TOPIC_FLOWS_PROMPT` flow join all end without trailing whitespace, so the join produces the same string the old `.format(...).strip()` did.
- **The `if section` filter** is doing real work via `generate_topic_flows_prompt()`, which returns `""` when no flows are enabled.

Nicole's `9e11313` is correct and complete: no raw `Site` writes remain in the test module, both converted tests wrap `site_update` in `captureOnCommitCallbacks(execute=True)`, and the kwargs match the service signature. The third case she mentioned, `test_blank_court_name_means_multi_court_mode`, needed no conversion because it asserts the default blank-court path and writes no Site at all.

Note for whoever merges: this was blocked only by the ruleset's `require_last_push_approval` — Nicole pushed the last commit and was the sole approver, so her approval couldn't carry it. Working as intended.

PR #843, opened 2026-08-31
docs: correct day-one inaccuracies in CLAUDE.md and corpus docs

### LINKED ISSUE #842: docs: four CLAUDE.md and corpus-doc inaccuracies a new dev hits on day one
### What needs to change?

Four items, all "a new dev reads this and it is wrong or absent." A new dev picks up the AI lane this week.

**1. The corpus location is documented three different ways, and none of them mentions there are two trees.**

- `CLAUDE.md:79` — corpus YAML lives in `litigant_portal/corpus/`
- `CLAUDE.md:178` — the corpus engine reads YAML from `litigant_portal/content/`
- `corpus-sources/README.md:5` and `:39` — author corpus in `litigant_portal/content/`

Each is individually true and together they mislead. `content/` is read by `app/topic_flow/registry.py` and renders the live public flow pages. `corpus/` is read by `app/selectors/corpus.py` and syncs to database rows. Both are live, both hand-maintained, both validated at startup by their own system check (`apps.py` registers `checks.corpus` and `topic_flow.checks`). All three places should say that two trees exist and what each serves.

Which tree retires, and when, is a separate open decision that sits under #179. Not this issue.

**2. `CLAUDE.md:30` gives the wrong precondition for the fast test suite.**

It says a focused test can be run "when `.tox/fast` already exists." Existence is not sufficient — a stale env exists and fails. `django-linear-migrations` arrived in #826 on the same day #480 documented the suite, so any env built before that dies on `ModuleNotFoundError: No module named 'django_linear_migrations'`. The instruction needs the recreate step: `tox -e fast --recreate` after a dependency change.

**3. Nothing documents how to pick up a dependency change in Docker.**

`make docker-down` is listed as "Stop containers," which is accurate, but `docker compose up` afterwards reuses the stale image. `make docker-up-build` is the target that rebuilds. Same root cause as item 2, different surface, and both bite on the first pull after a lockfile change.

**4. Two CLAUDE.md sections are written but unlanded.**

- **Review requests and CODEOWNERS** — CODEOWNERS puts every owner on every PR while the ruleset needs one approval, so pending reviewers are the normal state, not drift. Flag only on `assignees`, never `reviewRequests`.
- **There is one chat** — the prompt-backed chat is gone; the model-backed one is simply *chat*. Older issue titles still say "v2 chat" (#668, #715, #755–#762), which reads to a new dev as a live distinction.

### Why?

Every one of these cost real time this week. The corpus contradiction is the worst of them, because corpus is the first thing AI grounding work touches and editing the wrong tree fails silently — the page just does not change.

- [ ] `CLAUDE.md` and `corpus-sources/README.md` state that two corpus trees exist and what each serves
- [ ] `CLAUDE.md` names `tox -e fast --recreate` as the step after a dependency change
- [ ] `make docker-up-build` documented as the pick-up-dependency-changes path
- [ ] The CODEOWNERS review rule and the one-chat naming note are in `CLAUDE.md` on `main`

mitch-mitchel:
Four things a new dev reads in this repo and finds wrong or missing. A new dev picks up the AI lane this week, so these are worth correcting ahead of that.

The one that matters most is the corpus contradiction, because it is the only defect here that fails **silently**: `CLAUDE.md` named `litigant_portal/corpus/` as the corpus home in one place and `litigant_portal/content/` in another, while `corpus-sources/README.md` told authors to work in `content/`. Each statement was true and together they misled. Both trees are live, hand-maintained, and separately validated at startup — and editing one does not change the other, so authoring into the wrong tree produces no error and no visible change.

Also lands two `CLAUDE.md` sections that were written but never committed: the CODEOWNERS review-request rule, and the one-chat naming note.

Closes #842

## Test plan

Docs-only; no code paths touched. Each claim added is checkable:

- [ ] `make pre-commit` clean (prettier reflowed the new table on commit)
- [ ] `content/` renders the public pages — `app/topic_flow/registry.py` module docstring and `views/pages.py` importing `registry`
- [ ] `corpus/` syncs to the DB — `app/selectors/corpus.py:27` (`CORPUS_DIR`) and `manage sync_corpus`
- [ ] Both validated at startup — `app/apps.py` registers `checks.corpus` and `topic_flow.checks`
- [ ] `tox -e fast --recreate` fixes a stale `.tox/fast` (reproduced today: `ModuleNotFoundError: No module named 'django_linear_migrations'`, from #826 landing after the env was built)
- [ ] `make docker-up-build` picks up a lockfile change where `make docker-down` + `make docker` does not (same reproduction)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -27,7 +27,7 @@ Chat model is configurable via `CHAT_MODEL` env var (LiteLLM format, e.g. `opena
 
 **`make lint`, `make test`, `make pre-commit`** — sandbox restrictions prevent Claude from running these Docker- and Postgres-backed targets. Mitch runs the full lint/test workflow as part of his own process. A single post-commit mention is plenty; don't re-prompt about it across the commit/PR steps.
 
-**DB-free fast tests** — when `.tox/fast` already exists, Claude can run a focused test directly with `.tox/fast/bin/pytest <path>`. Use that suite for a real RED→GREEN cycle on non-DB units: write the focused test, run it and confirm that it fails for the expected reason, implement the change, then rerun it to green. Tests marked `postgres` require `make test` and Docker. The fast marker filter is not a complete database-isolation boundary: unmarked tests may still use Django's database, so run only focused tests already known to be DB-free through this path.
+**DB-free fast tests** — when `.tox/fast` exists _and_ is current, Claude can run a focused test directly with `.tox/fast/bin/pytest <path>`. **Existence is not enough:** the env is built from `uv.lock`, and a stale one fails at import (`ModuleNotFoundError`) rather than reporting a test failure. After any dependency change, `tox -e fast --recreate`. Use that suite for a real RED→GREEN cycle on non-DB units: write the focused test, run it and confirm that it fails for the expected reason, implement the change, then rerun it to green. Tests marked `postgres` require `make test` and Docker. The fast marker filter is not a complete database-isolation boundary: unmarked tests may still use Django's database, so run only focused tests already known to be DB-free through this path.
 
 ### Local Development (Docker)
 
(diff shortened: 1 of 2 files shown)


nicolearagao (approved):
LGTM

Issue #589, opened 2026-07-01
Docs drift + v1 orphan sweep: retire chat/ references after v2 chat lands


mitch-mitchel:
> [!WARNING]
> **Blocked — do not execute until:**
>
> 1. **v2 chat lands**, and
> 2. **Nathan's v1 code cleanup is done** (v1 removal is his; this issue does _not_ delete v1 code).
>
> Running the sweep before then would document a layout that's still mid-transition.

## What

Two coupled cleanups, both gated on the above.

**Docs** — `CLAUDE.md`'s "AI Chat Feature" + "Key Files" sections still describe the pre-move `chat/` app (the app was relocated to `litigant_portal/`; old code is in `archive/old_modules/chat/`):

- **Chat Architecture tree** roots at `chat/` with a `providers/` subtree (`base.py` / `ollama.py` / `factory.py`) — the app is now `litigant_portal/app/`, and the provider abstraction is gone entirely (LiteLLM replaced it, as the "LLM Provider" note already says). Services are at `litigant_portal/app/services/`; models and views are packages (`litigant_portal/app/models/`, `litigant_portal/app/views/`).
- **"How It Works" + "Chat Endpoints"** describe `POST /chat/send/` → `GET /chat/stream/`. Real routes (`litigant_portal/app/urls.py`) are `stream/`, `search/`, `chat-stream/` (chat_v2), etc. — there is no `/chat/send/`. The runtime flow actually changed, so this needs the views read, not a path swap.
- **Key Files table** points at `chat/` and `portal/views.py` — both moved.

**Orphan sweep** — after Nathan removes v1, sweep for leftover v1 artifacts: `archive/old_modules/chat/` (decide keep vs drop), stray root `chat/` pycache, and any remaining doc/config referencing the old paths.

## Why

`CLAUDE.md` is the repo's onboarding contract; a wrong architecture map misleads every new contributor (and Claude). Best done once, after the v1→v2 transition settles, so the docs match a stable v2 layout instead of a moving target.

## Definition of done

- `CLAUDE.md` Chat Architecture tree, "How It Works", endpoints, and Key Files all match `litigant_portal/app/urls.py` and the real v2 flow.
- No orphaned v1 files/paths remain — or `archive/` retention is a deliberate, noted decision.
- No overlap with `prompts-as-infra.md`, which was already synced under #340.

---

_Surfaced while doing #340 — the `prompts-as-infra.md` half is handled there; this is the larger, transition-gated remainder._

PR #845, opened 2026-09-01
test: add coverage for chat engine tool execution, thread endpoints, and upload serialization

### LINKED ISSUE #226: test: add coverage for upload_document, summarize_conversation, action_plan, ExtractionService
## Summary

Several views and services with real business logic have zero test coverage. These are the highest-priority gaps from the April 2026 test audit.

### Views (chat/views.py)
- **`upload_document`** — multiple error branches (no file, invalid PDF, extraction failure), completely untested
- **`summarize_conversation`** — missing messages, invalid JSON, too-few messages, agent unavailable, successful summary — all untested
- **`action_plan`** — context-building from `CaseInfo.data`, empty/populated states untested

### Services
- **`ExtractionService._parse_extraction_result()`** (chat/services/extraction_service.py) — substantial custom field-mapping logic, zero tests
- **`PDFService.extract_text()`** — pdfplumber integration path untested

### Agents (chat/agents/)
- **`Agent.messages_for_llm`** — `data` field stripping from tool messages
- **`Agent.execute_tool()`** — error handling (unknown tool, exception in tool)

## MEDIUM priority gaps

- `toast_messages` context processor — `error`→`danger` remapping
- `AnonymousSessionKeyMiddleware` — only tested indirectly
- `CaseInfo.__str__()` and `TimelineEvent.__str__()` — custom logic
- `stream()` view — `agent_name`/`topic` params, `KeyError` branch
- `chat_page()` — `topic_context` string-building

nicolearagao:
## What changed

**`_execute_tool` error normalization** (`test_execute_tool.py`, DB-free)
Unknown tool, invalid args, and a raising tool all come back as `ToolOutput` error results — an exception escaping the tool loop would kill the SSE stream mid-conversation. Also pins the current `extra="allow"` behavior: a hallucinated argument name is silently ignored, not rejected (see note below).

**Thread endpoint identity boundary** (`test_chat_thread_endpoints.py`, postgres)
`thread_list`, `message_list`, `thread_usage`, and `thread_delete` had zero direct tests. Each is a cross-identity ownership check (`chat_thread_get` scopes by identity and thread_type); a regression means one litigant reading another's conversation. Includes the superuser gate on `thread_usage`, and pins the delete cascade so the audit-log retention work (#745) changes it deliberately, not silently.

**Upload serialization contract** (`test_upload.py`, postgres)
`user_upload_serialize` and `user_upload_render_list` are the JSON shapes the chat frontend renders for attachments: payload contract, `is_image` flag, requested-order preservation, and the deleted-upload stub.

Closes #226 

## Test plan

---

### DIFF
--- a/litigant_portal/app/tests/test_chat_thread_endpoints.py
+++ b/litigant_portal/app/tests/test_chat_thread_endpoints.py
@@ -1,0 +1,257 @@
+"""Tests for the thread endpoints in views/chat_engine.py, through the
+assistant surface.
+
+Every endpoint is a cross-identity ownership boundary: chat_thread_get scopes
+by identity and thread_type, and a regression there means one litigant reading
+another's conversation. All tests are postgres-marked (`make test`).
+"""
+
+import pytest
+from django.contrib.auth import get_user_model
+from django.test import Client
+
+from litigant_portal.app.models import ChatMessage, ChatThread, UserIdentity
+from litigant_portal.app.services.chat_engine import chat_message_create
+from litigant_portal.app.services.user import user_identity_ensure
+
+pytestmark = [pytest.mark.postgres, pytest.mark.django_db]
+
+ASSISTANT_BASE = "/api/agents/assistant/"
+MODEL = "gpt-5-mini"
+
+
+def _client_identity():
+    """A client with an established anonymous session and its identity."""
+    client = Client()
+    client.get(ASSISTANT_BASE + "threads/")
+    identity = UserIdentity.objects.get(session_key=client.session.session_key)
+    return client, identity
+
+
+def _superuser_client_identity():
+    # Identity for an authenticated user is keyed by user, not session key,
+    # so login-time session rotation doesn't detach it.
+    user = get_user_model().objects.create_superuser(
+        username="root", password="pw"
+    )
+    client = Client()
+    client.force_login(user)
+    return client, user_identity_ensure(user=user)
+
+
+def _thread(identity, thread_type="user_chat", **kwargs):
+    return ChatThread.objects.create(
+        identity=identity, thread_type=thread_type, **kwargs
+    )
+
+
+def _message(thread, content, role="user", **kwargs):
+    return chat_message_create(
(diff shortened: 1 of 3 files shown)


mitch-mitchel (commented):
Three nits, none blocking. Nice set — the cross-identity 404s on both the read and the delete path are the highest-value part.

1. **`test_chat_thread_endpoints.py`** — `assert isinstance(payload["items"], list)` passes on an empty list, so it survives a regression that drops items entirely. `len(payload["items"]) == 1` costs nothing and still leaves contents to `test_chat_hidden.py`.

2. **`test_upload.py`** — `SimpleUploadedFile` is imported inside `_upload` while `user_upload_create` got hoisted to the top-level block in this same PR. Repo convention is top-level only (the locals at 256–300 are legacy); worth hoisting while the import block is already open.

3. **PR body** — "(see note below)" points at a code comment in `test_execute_tool.py`, not at anything in the body.

On `Closes #226`: it holds up, but the issue reads as though its whole list is still on the table. I walked all 12 of its bullets and left the disposition there — 8 died with the v1 chat retirement (#589), 3 already have coverage (`messages_for_llm`, `AnonymousSessionKeyMiddleware`, `toast_messages`), and this PR adds the last live one.

mitch-mitchel (dismissed):
good with non blocking comments

mitch-mitchel:
Nits 1 and 2 are fixed in e223967 — you're out Wed/Thu, so I pushed them rather than leave the PR parked.

- `len(payload["items"]) == 1` instead of the `isinstance` check. Confirmed against `chat_thread_render_items` first: one `kind: "user"` item per visible user message, and the test creates exactly one non-hidden, non-meta message.
- `SimpleUploadedFile` hoisted to the top-level block. Left the function-local imports in the older classes alone — that's a separate cleanup, not yours to carry here.

I couldn't run either file (both `postgres`-marked), so they want a `make test` before merge.

Nit 3 is the "(see note below)" in the description pointing at a code comment rather than anything in the body — left that one for you, it's your prose.

PR #853, opened 2026-09-02
docs(frontend): design principles and component rules

### LINKED ISSUE #852: Front-end principles doc and component rules in CLAUDE.md
### What needs to change?

- Publish `docs/wiki/frontend-design-principles.md`: why LP uses Atomic Design, where it costs us, what the stack doesn't give us and what we build instead, and the rules.
- Add the operative rules to `CLAUDE.md` inline, linking the doc for reasoning and tradeoffs.

### Why?

Unenforced prose doesn't reach outside contributors or agents that never open `CLAUDE.md`. That file is read every session; a referenced doc is read only when something decides to fetch it. So the rules live inline and the reasoning lives in the doc — the same split the permissions and WCAG sections already use.

- [ ] Principles doc published under `docs/wiki/`
- [ ] `CLAUDE.md` carries the operative rules inline and links the doc
- [ ] Extraction threshold consistent across the file (two templates, not three)

mitch-mitchel:
Records why LP uses Atomic Design, which had never been written down anywhere: Django and Cotton ship no component workbench, so the tiers plus `/style-guide/` stand in for Storybook, and the driver was WCAG AA and composed mobile-first design.

Adds the operative rules to `CLAUDE.md` inline with the reasoning and tradeoffs in the doc, lowers the bottom-up extraction threshold from three templates to two, and drops the `~/flp/CLAUDE.md` reference — a local file no contributor or CI run can read.

Closes #852

**Test plan:** docs only, no code paths touched. Confirm the relative link from `CLAUDE.md` to `docs/wiki/frontend-design-principles.md` resolves on GitHub, and that prettier is clean on both files.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -249,13 +249,21 @@ litigant_portal/app/templates/cotton/
 
 Style guide available at `/style-guide/` during development.
 
-Component & style discipline follows the org rule (`~/flp/CLAUDE.md`): compose from existing components first, extend with a prop when almost-right, new components/tokens only when nothing combines. LP paths: components in `litigant_portal/app/templates/cotton/`, theme tokens in `litigant_portal/app/src/main.css`. Check props (variants, sizes, `href`, `full_width`, `class` passthrough) before assuming a component can't do it.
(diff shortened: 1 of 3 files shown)


nicolearagao (approved):
LGTM, thank u

Issue #860, opened 2026-09-04
Artificial Intelligence (AI) and Platform Game Plan


Sinjhin:
### What problem does this solve?

This is my proposed direction for the AI and adjacent platform work. The architecture section is intentionally independent of the current issue list. The final section maps the proposal to current work so we can reuse what fits, change what does not, and see what is missing.

This is mostly for review at this stage, won't block me from getting started on implementation, but will need to be broken down into smaller tasks for actually doing this in digestable/reviewable chunks.

## Working priorities

These lists are intentionally high-level and movable. The details are in the architecture and issue sections below.

### Wednesday, September 9 Goal

- **AI Plan Solidified and Prioritized**
  - **issue/dependency mapping to current work**
- **Agent Ingress and Egress Contracts**
- **Agent Core Runtime**
- **Agent Redis Stream and Event Relay to Web**
- **Beginning Agent Tools**
- **Stretch Goal: Initial eval baseline**

### Wednesday, September 23 MVP Goal

- **Agent Run Persistence and `agent_` Data Model**
- **Full MVP Agent Runtime and Tools**
  - **Non-AI Fallbacks**
  - **Court and Topic Grounding**
  - **Opt-In Semantic Memory** (To be discussed)
  - **Conversation Compaction**
  - **~~Model Context Protocol Integration~~**
- **Document Ingestion/Vectorization and Semantic Vector Search/Retrieval**
- **MVP Observability and Auditability**
- **Safety Boundaries**
- **MVP Evaluation and Subject-Matter Review**
- **DoD Satisfied**

### Long-Horizon Goals

- **Expanded Evaluation, Observability, and Cost Controls**
- **Cross-Model and Cross-Harness Evaluation**
- **Agent2Agent Evaluation**
- **Runtime and Scaling Evaluation**

## Architecture direction

- **AI boundary and data ownership**
  - **Direction:** Put agent behavior behind a documented internal interface for requests, responses, tools, state, persistence, and failures that Django, workers, tests, and protocol adapters can share.
  - **Data:** Give new agent-owned models and tables an `agent_` prefix so their purpose is obvious. Continue using the shared fact model for user-confirmed case facts rather than creating agent-specific duplicates.
  - **Reasoning:** This makes the system easier to test, replace, audit, and explain without requiring a microservice split.

- **Agent execution and tools**
[...]

PR #862, opened 2026-09-08
860: First PR defining agent arch and interfaces

Sinjhin:
## What changed

Introduces a standalone `lp_agent` package and establishes the `LPAgent` interface for future agent execution. The contract covers submitting work, receiving updates and outcomes, handling user replies, and providing verified identity and court/topic context.

Defines a boundary between reusable agent behavior and application-specific services, with a thin `PortalAgent` entry point for Django. Packaging, Docker, and continuous integration support allow the core contract to be tested independently of Django, a database, or a model provider.

This PR provides contract scaffolding and validation. Runtime execution and concrete service adapters follow in subsequent PRs.

Part of #860.

## Test plan

- [x] Pre-commit checks on changed files and static typing checks passed.
- [x] `tox -e agent`: 57 contract tests passed without Django, a database, or provider credentials.
- [x] `make test`: application tests passed inside Docker.
- [x] Wheel and Docker image builds passed; the installed package imports outside the checkout without Django.

The checks above were run separately; the combined `make pre-commit` target remains unchecked.

---

### DIFF
--- a/docker/django/Dockerfile
+++ b/docker/django/Dockerfile
@@ -43,12 +43,13 @@ RUN TAILWIND_VERSION=${TAILWIND_VERSION} \
 # Stage 4: Application code
 # -----------------------------------------------------------------------------
 COPY litigant_portal litigant_portal
-COPY lp_agent lp_agent
 
 RUN /tmp/tailwindcss \
     -i /app/litigant_portal/app/src/main.css \
     -o /app/litigant_portal/app/static/css/main.built.css \
     --minify
+
+COPY lp_agent lp_agent
 
 RUN --mount=type=cache,target=/root/.cache/uv \
     uv pip install --no-deps --python /opt/venv/bin/python -e .
--- a/docker/django/Dockerfile
+++ b/docker/django/Dockerfile
@@ -43,6 +43,7 @@ RUN TAILWIND_VERSION=${TAILWIND_VERSION} \
 # Stage 4: Application code
 # -----------------------------------------------------------------------------
 COPY litigant_portal litigant_portal
+COPY lp_agent lp_agent
 
 RUN /tmp/tailwindcss \
     -i /app/litigant_portal/app/src/main.css \
--- a/litigant_portal/agent.py
+++ b/litigant_portal/agent.py
@@ -43,7 +43,9 @@ class PortalAgent(LPAgent):
             )
             ScopeSelection(court=court, topic=topic)
         except ValidationError as exc:
-            raise AgentValidationError.from_validation_error(exc) from exc
+            raise AgentValidationError.from_validation_error(
+                exc, models=(AgentConfiguration, ScopeSelection)
+            ) from exc
         if identity is None:
             raise AgentValidationError("a host-verified identity is required")
         raise NotImplementedError(
--- a/litigant_portal/agent.py
+++ b/litigant_portal/agent.py
@@ -1,0 +1,52 @@
+"""
+Host entry point for the new agent; Django adapter wiring follows in PR2.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from pydantic import ValidationError
+
+from lp_agent import AgentValidationError, LPAgent, RunLimits
+from lp_agent.types import (
+    AgentConfiguration,
+    InterruptBehavior,
+    Runtime,
+    ScopeSelection,
+)
+
+if TYPE_CHECKING:
+    from litigant_portal.app.models import UserIdentity
+
+
+class PortalAgent(LPAgent):
+    """
+    Wire host-verified identity, application adapters, and portal defaults.
+    """
+
+    def __init__(
+        self,
+        *,
+        identity: UserIdentity,
+        court: str | None = None,
+        topic: str | None = None,
+        runtime: Runtime = "Workers",
+        interrupt_behavior: InterruptBehavior = "reject",
+        limits: RunLimits | None = None,
+    ) -> None:
+        try:
+            AgentConfiguration(
+                runtime=runtime,
+                interrupt_behavior=interrupt_behavior,
+                limits=RunLimits() if limits is None else limits,
+            )
+            ScopeSelection(court=court, topic=topic)
+        except ValidationError as exc:
+            raise AgentValidationError.from_validation_error(exc) from exc
+        if identity is None:
+            raise AgentValidationError("a host-verified identity is required")
+        raise NotImplementedError(
+            "Django environment construction is not implemented; "
+            "PortalAgent adapter wiring follows in PR2."
+        )
--- a/lp_agent/errors.py
+++ b/lp_agent/errors.py
@@ -4,7 +4,55 @@ Errors raised before a run or input is accepted.
 
 from typing import Self
 
-from pydantic import ValidationError
+from pydantic import BaseModel, ValidationError
+
+_ERROR_MESSAGES = {
+    "extra_forbidden": "Unexpected field",
+    "missing": "Required field",
+    "string_type": "Must be text",
+    "int_type": "Must be an integer",
+    "float_type": "Must be a number",
+    "bool_type": "Must be a boolean",
+    "tuple_type": "Must be a sequence",
+    "list_type": "Must be a sequence",
+    "dict_type": "Must be an object",
+    "greater_than": "Value is too small",
+    "greater_than_equal": "Value is too small",
+    "less_than": "Value is too large",
+    "less_than_equal": "Value is too large",
+    "finite_number": "Must be a finite number",
+    "literal_error": "Unsupported option",
+    "union_tag_invalid": "Unsupported option",
+    "union_tag_not_found": "Missing option type",
+}
+
+
+def _safe_location(location: tuple, schema: dict) -> str:
+    """
+    Follow declared properties and sequence indices, never dictionary keys.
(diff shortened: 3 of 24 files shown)


lp_agent/errors.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @classmethod
+    def from_validation_error(cls, error: ValidationError) -> Self:
+        """
+        Describe invalid fields without including the submitted values.
+        """
+        problems = []
+        for item in error.errors(
+            include_url=False, include_context=False, include_input=False
+        ):
```
nicolearagao, line 86: It correctly hides the values the user sent, but it still prints the field names. As we are using extra="forbid", an unknown field will become part of the error. 
For unknown fields, we could replace the location with something generic. what do you think ?

tox.ini
```diff
@@ -28,3 +28,15 @@ setenv =
...
     POSTGRES_PASSWORD = postgres
 commands = pytest -m "not postgres" {posargs:--verbosity=1}
+
+[testenv:agent]
+# Exercise the core without installing Django or initializing its settings.
+runner = uv-venv-runner
+package = skip
+extras =
+deps =
+    pydantic>=2.0
```
nicolearagao: The agent env declares pydantic>=2.0, but types.py uses JsonValue (2.5+) and PEP 695 type aliases in fields (2.6+). Works today because resolution picks the newest. Should be >=2.6.

tox.ini
nicolearagao: The new agent env is not in envlist (still py{313}), and make test runs plain tox via docker/django/entrypoint.sh (exec tox "${@:2}"), so the isolation env never runs locally. Only the new CI step (tox run -e agent) exercises it.
the lp_agent test files do run under make test, because testpaths in pyproject.toml now includes lp_agent/tests. But that run happens with Django installed, which is exactly the condition this env exists to remove, right? So if someone adds a Django import to lp_agent, make test stays green and the regression only shows up in CI.

lp_agent/types.py
```diff
@@ -0,0 +1,372 @@
...
+    model_config = ConfigDict(extra="forbid", frozen=True, allow_inf_nan=False)
+
+
+class RunLimits(ContractModel):
+    """
+    Cumulative budgets, excluding queue time and time waiting for input.
+    """
+
+    max_steps: Annotated[StrictInt, Field(gt=0)] = 30
+    max_active_seconds: Annotated[float, Field(gt=0, strict=True)] = 300
```
nicolearagao: max_active_seconds is declared strict float but defaults to the int 300, and Pydantic doesn't validate defaults. RunLimits() and RunLimits(max_active_seconds=300.0) serialize differently (300 vs 300.0), which will matter once PR2 persists and compares configurations. Fix: = 300.0 or validate_default=True.

litigant_portal/agent.py
```diff
@@ -0,0 +1,52 @@
...
+        try:
+            AgentConfiguration(
+                runtime=runtime,
+                interrupt_behavior=interrupt_behavior,
+                limits=RunLimits() if limits is None else limits,
+            )
+            ScopeSelection(court=court, topic=topic)
+        except ValidationError as exc:
+            raise AgentValidationError.from_validation_error(exc) from exc
+        if identity is None:
```
nicolearagao, line 49: The identity guard only rejects the literal None, and it runs after the config/scope validation. object(), "", an AnonymousUser, or an unsaved UserIdentity (pk=None, which later becomes the string "None" in AccessContext) all pass as "host-verified identity". Since this is the security anchor of the contract, I'd validate it first and for real.

litigant_portal/agent.py
```diff
@@ -0,0 +1,52 @@
...
+    ) -> None:
+        try:
+            AgentConfiguration(
+                runtime=runtime,
+                interrupt_behavior=interrupt_behavior,
+                limits=RunLimits() if limits is None else limits,
+            )
+            ScopeSelection(court=court, topic=topic)
+        except ValidationError as exc:
+            raise AgentValidationError.from_validation_error(exc) from exc
```
nicolearagao: This block duplicates LPAgent.__init__'s validation and discards the result, and super().__init__() is never called, so _configuration/_environment stay unset. Fine while everything raises NotImplementedError, but it's a trap for PR2 (instant AttributeError when the raise is removed) and the two copies will drift. Sharing one validation path would remove the trap.

lp_agent/types.py
```diff
@@ -0,0 +1,372 @@
...
+        Associate tool results with calls made by assistant messages.
+        """
+        if (self.role == "tool") != (self.tool_call_id is not None):
+            raise ValueError("tool_call_id is required only for tool messages")
+        if self.tool_calls and self.role != "assistant":
+            raise ValueError("only assistant messages may contain tool calls")
+        return self
+
+
+class ToolDefinition(ContractModel):
```
nicolearagao: ToolDefinition here is flat (name/description/input_schema), but our audit system fingerprints tool schemas in the OpenAI-nested shape: prompt_artifact_content_hash (services/chat_engine.py:41-54) SHA-256s the nested {"type":"function","function":{...}} schemas and stores them on PromptArtifact.tool_schemas. That hash is how we later prove "these are the exact instructions the model had."

If the new engine records tools in this flat shape, the identical tool gets a different hash than it has today, so the audit trail splits at the engine transition: an auditor sees two different artifacts for instructions that never changed.
can we pick one canonical shape for hashing before PR2? Either reuse the nested shape here, or keep the flat contract and convert to nested at hashing time. No preference on which, as long as both engines fingerprint the same representation.

docker/django/Dockerfile
```diff
@@ -43,6 +43,7 @@ RUN TAILWIND_VERSION=${TAILWIND_VERSION} \
 # Stage 4: Application code
 # -----------------------------------------------------------------------------
 COPY litigant_portal litigant_portal
+COPY lp_agent lp_agent
```
nicolearagao: COPY lp_agent sits above the Tailwind RUN, so every pure-Python agent change rebuilds the CSS. Moving the COPY below the Tailwind step (before the editable install) restores the cache boundary, and PR2/PR3 will feel the difference.

nicolearagao (commented):
Nice. The scope discipline is spot on: every entry point honestly raises NotImplementedError, the validation is real, and the import-isolation test is a genuinely good idea. I'm on board with the direction. Most of my comments are nits, but fixing now could simplify things down the road.

docker/django/Dockerfile
Sinjhin: Resolved with: [3da096a](https://github.com/freelawproject/litigant-portal/pull/862/commits/3da096a18489e837420662330437d975d900e0ab)

lp_agent/errors.py
Sinjhin: Resolved with: [3da096a](https://github.com/freelawproject/litigant-portal/pull/862/commits/3da096a18489e837420662330437d975d900e0ab)

I added a bit of defense here: an unexpected key under `limits` now gives `limits: Unexpected field` without repeating the key. These are errors from the programmatic interface; browser input doesn’t currently reach this code.

tox.ini
Sinjhin: Resolved with: [3da096a](https://github.com/freelawproject/litigant-portal/pull/862/commits/3da096a18489e837420662330437d975d900e0ab)

I bumped this to `2.10.0` because `2.8.2` and `2.9.0` still failed to load our types on Python 3.13.

tox.ini
Sinjhin: Resolved with: [3da096a](https://github.com/freelawproject/litigant-portal/pull/862/commits/3da096a18489e837420662330437d975d900e0ab)

`agent` is now in the default tox environments, and its explicit CI step remains. The import blocker also exercises the new model contracts.

lp_agent/types.py
Sinjhin: Resolved with: [3da096a](https://github.com/freelawproject/litigant-portal/pull/862/commits/3da096a18489e837420662330437d975d900e0ab)

litigant_portal/agent.py
Sinjhin: I’m leaving this for PR2, when the constructor will actually build an agent. Right now it always raises `NotImplementedError`, so it can’t use an unchecked identity.

litigant_portal/agent.py
Sinjhin: This belongs with the constructor wiring in PR2 too, when there’s a real environment to pass to `super().__init__()`. Both paths already use `AgentConfiguration`, so the validation rules themselves are shared.

lp_agent/types.py
Sinjhin: Resolved with: [3da096a](https://github.com/freelawproject/litigant-portal/pull/862/commits/3da096a18489e837420662330437d975d900e0ab)

I took this a bit further as discussed on Slack and used OpenAI’s Responses format for the new model interface. Since we’re not live yet, we can start with a fixed, versioned hash format without preserving hashes from the old format. Note that I have this essentially "vendored" off the spec, though if we ever see a need we can pull it from the OpenAI SDK. Same with the provider adapters in PR2 and the LiteLLM lib.

litigant_portal/agent.py
nicolearagao: that's fine

nicolearagao (approved):
LGTM, thanks for addressing the changes

Issue #777, opened 2026-08-06
Org proposal: adopt issue fields as the source of truth for priority and effort


mitch-mitchel:
**Decision needed from org owners and team leads.** Issue fields are configured org-wide, so this affects every team's repos and boards. Nothing changes until it's agreed.

## The principle

**The issue is the source of truth. Boards read from it.** An issue field's value lives on the issue and is the same in every project that issue belongs to. A project field lives on the project item, so the same issue can carry different values on different boards, with nothing reconciling them.

Boards keep everything they need: they group, filter, sort and chart by an issue field natively, and an edit made in a board syncs back to the issue. Issues also become searchable by value — `field.priority:P1`, `field.effort:S`.

## The ask

The org already has issue fields enabled, carrying GitHub's untouched defaults:

| Field | Today | Requested |
| --- | --- | --- |
| Priority | Urgent / High / Medium / Low | `P0` / `P1` / `P2` / `P3` |
| Effort | High / Medium / Low | `XS` / `S` / `M` / `L` / `XL` |
| Visibility (both) | Organization-only | **Public** |
| Start date / Target date | unused | keep or delete |

## Impact

**The migration is one value.** Across the last 50 issues in each of litigant-portal, courtlistener and infrastructure, exactly one carries an issue-field value: courtlistener#7742, `Priority=Low`. Nothing else in the org has adopted these fields, so re-optioning them is close to a no-op.

**Nothing is forced on any team.** A board opts in by adding the field as a column, once. Teams that don't add it see no change. Issue fields and project fields coexist, so each team retires its own Priority/Size fields and its `P*` / `size: *` labels when it's ready.

**Public means public.** Priority and effort on public repos become readable by anyone. It is also required: organization-only fields aren't available on public projects at all, and flipping a field back to organization-only automatically removes it from any public project using it.

**One scale for everyone.** Options are org-wide, so teams converge on `P0`–`P3` and `XS`–`XL` or don't use the field. Some boards run their own variants today (separate `Size`, `PR Review Size`, `The actual size was...`).

**Two gaps to expect.** Pull requests and draft issues carry no issue-field values, so boards holding them show blank cells. And any tooling reading priority or size from labels needs updating before those labels retire.

## Open for the group

- Adopt `P0`–`P3` and `XS`–`XL` as the org-wide scales, or leave a team's variant out
- Public, or stay organization-only and forgo public projects
- Keep or delete the unused `Start date` / `Target date`

## Reference

[...]

Issue #75, opened 2026-01-12
LP Simple Auth


mitch-mitchel:
- login/logout modal
- user badge/login status

Issue #61, opened 2026-01-07
Move AlpineJS (for sourcemaps), AlpineJS minified and Tailwind local to repo


mitch-mitchel:
Had issues with AlpineJS not loading from the CDN, need to move it and Tailwind to local files. The prod release will use those local files to build smaller artifacts, but for dev we should have all of them local as full sources.

PR #863, opened 2026-09-08
Document how work is tracked and where priority and size live

### LINKED ISSUE #776: Consolidate work-tracking surfaces: leave board #61, document how work is tracked
## What

Priority and size are currently recorded in four places on the same issue. #587 is the specimen: repo label `P1`, org issue field Priority `High`, board #75 Priority `P1`, board #61 Priority unset. The label says `size: S` while the org Effort field says `Medium` — already drifted, and nothing reconciles them. The native issue Type duplicates the `task` / `bug` / `enhancement` labels on top of that.

This issue takes the steps that need no org coordination.

**1. Leave board #61.** 112 LP issues sit on the Web Team board — 45 General Backlog, 19 Someday, 16 To Do, 13 with no status at all. Nobody grooms them and their Priority/Size fields are empty. LP already reports its sprint from #75, so removing LP items from #61 loses nothing.

**2. Labels stay canonical; #75 fields are a one-way projection.** Priority and size are edited as labels on the issue, by anyone with write access. `board-sync.py` pushes label → board field. The board fields can't simply be deleted, because a project cannot group by label — that limit is why they exist. Nobody hand-edits them.

**3. Write down how work is tracked.** Today this lives only in agent instructions outside the repo:

- OSS standards top to bottom. Assign on start of work, active iteration included. Unassigned is legitimate everywhere.
- Iterations are numbered and dated. No letter/artist crosswalk to Web Team sprint names — report LP's iteration and its date window at the shared retro.
- Carry-over happens at planning. Unfinished work returns to Backlog, never auto-rolled into a future iteration.
- Iteration field = commitment. Status = pipeline position.
  - Backlog: not committed to an iteration, no Iteration set, size and priority **not** required
  - Ready: committed to the current iteration, not started — Iteration, size and priority set
  - Size and priority are the gate to *leave* Backlog, not to enter it
  - Work placed in a future iteration stays Backlog until that iteration opens
  - Non-work items (decision records, research, `question`) live in Backlog permanently and unsized

## Why

CONTRIBUTING.md covers issue → branch → PR → review. `docs/wiki/issue-conventions.md` covers templates and labels. `docs/wiki/definition-of-done.md` covers the quality bar. None of them say how the board works, what Backlog vs Ready means, or when work gets assigned. For a public repo that belongs in the docs, not in a private instruction file.

## Definition of done

- [x] LP items removed from board #61
- [x] CONTRIBUTING.md gains a "How work is tracked" section: board, iterations, statuses, assign-on-start
- [x] `docs/wiki/issue-conventions.md` states that priority and size are edited as labels, and the board field is a projection
- [x] `CLAUDE.md` points at both instead of restating them
- [x] `board-sync.py` write path is one-way, label → field
- [x] `/board-audit` skips unsized non-work items instead of flagging them

mitch-mitchel:
Board usage had no contributor-facing home — it lived only in instruction files outside the repo. `CONTRIBUTING.md` gains a "How work is tracked" section covering the board, assignment, iterations and the Backlog through Done statuses; `docs/wiki/issue-conventions.md` states that the `P*` and `size:` labels are the source of truth; `CLAUDE.md` points at both instead of restating them.

The board mirror exists because a project board can't group or sort by label, not as a second place to record the value. Org-wide issue fields as the eventual single surface are tracked in #777.

Two things were corrected against the original August write-up, since the board moved while it sat: the letter/artist sprint crosswalk does still exist (kept outside the repo, read by `/board-audit`), and carry-over is additive — a closing iteration's committed total is recorded before anything moves.

Closes #776

## Test plan

Documentation only, no code paths touched. Verified against live state rather than the August assumptions:

- Board #75 is public, so the CONTRIBUTING link is readable by non-members
- Board #61 has zero litigant-portal items
- `board-sync.py`'s write path is one-way, label → field
- `/board-audit` exempts unsized Backlog and never-sized non-work items
- Anchor links between the three files resolve to real headings

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,13 +155,18 @@ EOF
 
 Field ids are validated against the chosen template — a section that doesn't match (wrong field or wrong template) warns and would render blank. The label is applied by the template; set assignee/priority/size in the browser. See `scripts/file_issue.py` for the full format and accepted type aliases.
 
-## Sprint mapping
+## Work tracking
 
-When someone references a sprint by its web-team letter/artist name ("the Ed Sheeran sprint", "Sprint F"), translate to the matching LP Iteration on board #75 via the sprint-map crosswalk (JI-team record kept outside the repo; the board/vault tooling holds the current copy), then pull the work from #75 + git. This crosswalk is LP-specific — other JI repos don't necessarily align with the web-team retro, so it lives here, not in org-level instructions.
+Contributor-facing docs are canonical — don't restate them here:
+
+- **Board, iterations, statuses, assignment:** [CONTRIBUTING.md](CONTRIBUTING.md#how-work-is-tracked)
+- **Where priority and size live, and why the board mirrors them:** [docs/wiki/issue-conventions.md](docs/wiki/issue-conventions.md#where-priority-and-size-live)
+
+**Sprint mapping.** LP iterations are numbered and dated; the web team names its sprints by letter/artist. When someone references "Sprint F" or an artist name, translate to the matching LP Iteration through the sprint-map crosswalk (JI-team record kept outside the repo; the board/vault tooling holds the current copy), then pull the work from #75 + git. The crosswalk is LP-specific, so it lives here rather than at the org level.
 
 ## Sizing & estimation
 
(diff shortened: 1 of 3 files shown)


nicolearagao (approved):
LGTM

PR #864, opened 2026-09-09
feat(content): restore name and county questions

### LINKED ISSUE #803: Restore name/county guided-flow questions, protected the same way as publication_date

The ND name-change guided flow currently asks for only one fact (publication_date) — name and county were removed by #621 over a shared-terminal privacy concern (a stranger's leftover answer showing up for the next person on the same public computer). Without them, the docassemble prefill work (#531/#798) can only ever save a litigant from retyping one date; everything else on the real forms still needs manual entry, which makes the feature barely useful in practice. Approved by Jessica as part of the #798 guest/anonymous prefill scope decision.

Restore the name and county fact_gather questions (previously built and closed under #542, reference commit 9563099), applying the same protection already used for publication_date: collect the answer, use it immediately for that user, but never echo it back into the form or the recap screen, so a later visitor on a shared terminal never sees it (extend _NEVER_PREFILL in litigant_portal/app/topic_flow/renderer.py to cover these fields too).
Applies to both ND name-change tracks (standard and waiver). The same pattern generalizes to other topics (eviction, civil traffic) later — no new plumbing needed per field, just corpus content.

_No response_

nicolearagao:
## What changed

The two ND adult name change flows ask for first, middle, and last name plus county again, on both the standard and waiver tracks. #621 pruned these because their answers only echoed back into the summary and gave a shared terminal something to leak; they return now that the docassemble prefill has a real use for them, so the litigant does not retype what they already typed.

They come back under the shared-terminal protection rather than around it: all four ids join name_change_publication_date in the renderer's _NEVER_PREFILL, so the answers are stored and sent downstream but never rendered into the form or the recap. A litigant sees blank fields after saving, and the help text on the first name says so.

The questions are optional, unlike their glossary entries. A partly answered flow is a supported handoff case, and requiring them would block someone who only came for the publication deadline.

A new DB-free guard pins the two sides together: each flow asks all four questions, and each id is protected. The renderer's never-prefill tests now run over the whole protected set instead of the publication date alone.

Closes #803

## Test plan

- [x] manual checks 
Screenshot From 2026-09-09 17-00-12

Screenshot From 2026-09-09 17-00-03

---

### DIFF
--- a/litigant_portal/app/templates/cotton/molecules/toast_container.html
+++ b/litigant_portal/app/templates/cotton/molecules/toast_container.html
@@ -3,7 +3,8 @@
 {% if messages %}
   <div class="fixed top-4 left-1/2 -translate-x-1/2 z-50 w-full max-w-md px-4">
     {% for message in messages %}
-      <c-atoms.auto-dismiss>
+      {# 5s, not the atom's 1.5s default: flash messages are full sentences #}
+      <c-atoms.auto-dismiss timeout="5000">
       <c-atoms.alert variant="{{ message.variant }}" class="mb-2 shadow-lg">{{ message.text }}</c-atoms.alert>
       </c-atoms.auto-dismiss>
     {% endfor %}
--- a/litigant_portal/app/templates/pages/style_guide.html
+++ b/litigant_portal/app/templates/pages/style_guide.html
@@ -1035,7 +1035,7 @@
           <section id="toast-container" class="mb-12" aria-labelledby="toast-container-heading">
             <h3 id="toast-container-heading" class="mb-4">Toast Container</h3>
             <p class="text-greyscale-600 mb-6">Renders Django flash messages as auto-dismissing toast overlays. Included in <code>base.html</code>.</p>
-            <p class="text-sm text-greyscale-500 mb-4">Composes <code>c-atoms.auto-dismiss</code> and <code>c-atoms.alert</code>. No props — reads <code>messages</code> from template context.</p>
+            <p class="text-sm text-greyscale-500 mb-4">Composes <code>c-atoms.auto-dismiss</code> (at a 5s timeout, so sentence-length messages stay readable) and <code>c-atoms.alert</code>. No props — reads <code>messages</code> from template context.</p>
           </section>
           <section id="user-menu" class="mb-12" aria-labelledby="user-menu-heading">
             <h3 id="user-menu-heading" class="mb-4">User Menu</h3>
--- a/litigant_portal/app/tests/test_content_never_prefill.py
+++ b/litigant_portal/app/tests/test_content_never_prefill.py
@@ -1,0 +1,42 @@
+"""Drift guard: the name/county questions real content asks never echo back.
+
+The ND name-change flows collect them for the docassemble prefill, not to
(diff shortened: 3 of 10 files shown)


nicolearagao:
> **Additional findings** (outside the current diff — GitHub can't attach inline comments there):
> 
> * 🔴 `litigant_portal/app/views/pages.py` — Because NEVER_PREFILL fields (renderer.py:116) always render blank, resubmitting a fact_gather form that mixes them with other answers silently clears any sibling NEVER_PREFILL field the user didn't retype: a blank optional field maps to None here and update_or_create overwrites the stored value. The new 'your_information' section has 4 such fields in one , so returning later to add just one wipes the others already saved, defeating the docassemble-prefill goal this PR restores them for. Fix: don't persist None for a NEVER_PREFILL field on a resubmit unless the user explicitly signals clearing (e.g. skip writing when blank and an answer already exists, or use a distinct 'clear' control) rather than treating every blank as an intentional clear. [also at: litigant_portal/app/topic_flow/renderer.py:103 - Restoring first_name/middle_name/last_name/county as one multi-field NEVER_PREFILL section means editing any single…]
>   Extended reasoning...
>   User fills first_name/middle_name/last_name/county in the 'your_information' section and saves; renderer.py's NEVER_PREFILL set (line 97-103) blanks all four values on the next GET regardless of what's stored. Believing the section is unfilled, or just returning to add one missing field, the user retypes only one field and leaves the other three blank, then clicks Save. pages.py:96-100 computes valid[qid] = None for every blank optional field submitted (the docstring at line 94-95 documents this as intentional for a field whose blank means 'clear'), and variable_answer_set (services/topic_flow.py:109) does update_or_create with that None, overwriting the previously stored answer. All three untouched-but-previously-answered fields are now null in the DB, silently, even though the save 'succeeded' and shows the generic privacy-saved toast. This is a new failure mode: before this diff, no multi-field NEVER_PREFILL section existed, so a blank…
>   **Verification:** normal. This change introduces silent data loss. Chain (all verified): - New `your_information` fact_gather section (adult-name-change-standard.yml / -waiver.yml) holds four OPTIONAL fields first_name, middle_name, last_name, county in ONE form. flow_section_fact_gather.html:7 renders exactly one  per section with one input per question; none of the four have `required: true`. -… | normal.…

I am aware of this side effect and there's a fix for it n my next PR.

mitch-mitchel (approved):
Nice work — the `NEVER_PREFILL` widening, the drift guard pinning content against the protected set, and the blank-after-save toast all read well. Checked the naming against #869 and the explicit `interview_prefill` mapping answers it.

Tiny non-blocking nit: only `first_name` carries the "your current legal name" help text, so on a name-change form the middle and last fields inherit that context rather than stating it. Worth a word on each, or on the section heading.

nicolearagao:
> Nice work — the `NEVER_PREFILL` widening, the drift guard pinning content against the protected set, and the blank-after-save toast all read well. Checked the naming against #869 and the explicit `interview_prefill` mapping answers it.
> 
> Tiny non-blocking nit: only `first_name` carries the "your current legal name" help text, so on a name-change form the middle and last fields inherit that context rather than stating it. Worth a word on each, or on the section heading.

hey @mitch-mitchel , thanks for your review. I will merge this PR and include a commit with the nit above in my next PR.

Issue #805, opened 2026-08-17
Decide and implement mitigation for multi_user's unencrypted docassemble storage


nicolearagao:
### What needs to change?

multi_user = True is required for the push+resume prefill handoff to work (confirmed in #798), but it means litigant answers sit unencrypted in docassemble's database — readable by anyone with server or database access, not just the litigant. Existing hosting prot

### Why?

This is a new, distinct security gap introduced by making prefill work, not something already mitigated elsewhere. Leaning toward deleting the docassemble session shortly after the packet downloads, limiting how long the plaintext data exists — but this needs team's review to confirm it's sufficient, or to propose something stronger, before this ships to real litigants.

### Definition of done (optional)

- [ ] team reviews and confirms (or proposes an alternative to) the short-lived-session mitigation
- [ ] Approach implemented before the push+resume handoff ships

Issue #798, opened 2026-08-14
Investigate docassemble prefill options for the packet handoff


nicolearagao:
### What needs to change?

Research how an external app can start a docassemble interview that is already filled with known answers, so litigants don't retype what they already entered in the Topic Flow (#531, Phase 2).

Questions to answer:
- Does docassemble have an API that accepts a dictionary of answers?
- How does the user's browser get attached to a session created by our server?
- How do we keep personal data out of URLs (query strings, logs, browser history)?
- What server configuration does each option need (API keys, interview settings, minimum version)?

### Why?

Prefill is the main Phase 2 deliverable of #531. Before creating the build sub-issues, we need to know which handoff mechanism docassemble supports and pick one, since the choice shapes the LP endpoint, the interview YAML, and the server setup.

### Definition of done (optional)

- [ ] The viable mechanisms are written up with pros and cons (push: API session + one-time resume URL, vs. pull: token in the URL + interview fetches from an LP endpoint)
- [ ] Open questions listed for the design conversation (security tradeoff of multi_user = True, API key ownership, publication_date privacy rule, version pin)
- [ ] A recommendation posted on #531 so the build sub-issues can be filed

nicolearagao:
**Main finding: docassemble supports this.** There is no way to POST an answers dictionary directly to an interview page, so the choice is between two supported patterns.

## Option A: Push (LP calls the docassemble API)

This is docassemble's documented pattern for exactly our case.

1. `GET /api/session/new?i=<interview>` creates a session server-to-server.
2. `POST /api/session` with a JSON `variables` dictionary fills in the answers. Each key/value becomes an interview variable. ISO date strings become docassemble dates automatically.
3. `POST /api/resume_url` (with `one_time=1` and a short expiry) returns a one-time link like `/launch?c=<random-code>`. LP redirects the user's browser there. The code is opaque: no session ID and no personal data ever appear in a URL.

Docs:

- API endpoints (session/new, set variables, resume_url): https://docassemble.org/docs/api.html
- Security notes (why keys go in headers, not URLs): https://docassemble.org/docs/security.html

Requirements:

- A docassemble **API key**, sent in the `X-API-Key` header. Keys are per-user; an admin can scope them to specific permissions and restrict by IP.
- The interview must set **`multi_user = True`**, because an anonymous browser cannot hold the session's encryption secret. This turns off server-side encryption for that session's answers.
[...]

PR #869, opened 2026-09-09
feat(topic_flow): prefill the docassemble interview from guided answers

### LINKED ISSUE #804: Build the push + resume handoff: guided-flow answers to prefilled docassemble interview

Topic Flow's packet output already links out to a docassemble interview ("Fill out your forms" button, #543), but the litigant has to retype everything from scratch once there — the guided flow's own answers aren't carried over. #798's research confirmed the mechanism works: create a docassemble session via its API, push the guided flow's answers into it, generate a one-time resume link, and redirect the litigant there instead of the plain interview link.

Wire the packet output's "Fill out your forms" handoff to:
1. Call docassemble's /api/session/new for the target interview
2. POST /api/session with the guided flow's current AnswerStore values as variables (per the Topic Flow to docassemble contract in docs/docassemble.md)
3. POST /api/resume_url with one_time: 1 for a single-use link
4. Redirect the litigant's browser to that link instead of the static interview URL
Requires the target interview to set multi_user = True (confirmed necessary in #798 — without it, a resume link can't be retrieved from the litigant's own browser). Requires a docassemble API key provisioned for LP's server (owner/process TBD, tracked as an open item, not blocking this issue).
Depends on the name/county restoration issue above (so there's more than publication_date to send). Naming: use whichever filing_county/residence_county decision lands; default to residence_county if unresolved when this is picked up, since that's what the live interviews already use.

_No response_

nicolearagao:
## What changed

Closes #804.

The packet section's "Fill out your forms" button now hands docassemble the
litigant's confirmed answers instead of a cold start: it posts to a new LP
endpoint that creates a session over the docassemble API, pushes the answers
as interview variables, and redirects onto a one-time resume link. The
litigant lands inside the interview with those questions already answered.

One commit per piece, in dependency order:

- **`multi_user` in both ND interviews** (`a16b03c`). Required from an
  `initial` code block or the resume link cannot decrypt the session (#798).
  The QA playground is a manual upload, so committing this is not deploying it.
- **Blank protected fields stop erasing answers** (`a215546`). A
  `NEVER_PREFILL` field renders blank whatever is stored, so the handler read
  every re-save as "erase this". That was live data loss on main (re-saving
  the publication section destroyed the stored date and its 30-day deadline);
  the name/county questions would have multiplied it by four. Includes the
  `reviewed_only` filter on `variable_answer_map` that the payload reads
  through.
- **The docassemble client** (`ba148f4`). `session/new` → `session`
  (variables, `question: 0`) → `resume_url` (`one_time: 1`, 10-minute
  expiry). Key in a header, answers in POST bodies, nothing in a URL.
  `DOCASSEMBLE_BASE_URL` lets a local bench answer for the QA URL the
  content carries.
- **The name mapping** (`5dc4b0f`). Only 3 of the interview's 19 variables
  share our glossary names, so each flow's packet section carries an
  `interview_prefill` map, validated by the schema (interview names must be
  plain identifiers, since docassemble executes them as assignments) and the
  loader (keys must be that flow's question ids). A CI drift guard parses the
  versioned interviews: mapped variables must exist, checkboxes are never
  mapped, choice values must agree — a preset variable skips interview
  validation, so drift prints wrong values on court forms instead of failing.
- **The endpoint and button swap** (`ab4d477`). POST because session creation
  is a side effect. Only reviewed answers are sent: prefilled variables skip
  their questions, so confirmation has to happen on our side, which is the
  trust boundary the future AI fact-writing relies on. Any
  `DocassembleError` falls back to the plain interview link — no key,
  docassemble down, timeout — so the button never dies, and
  `DOCASSEMBLE_PUBLIC_URL` rewrites resume links minted against an internal
  address.
- **Hardening after review** (`1a29bbe`–`2fae9bb`): a CI guard that every
  handoff interview sets `multi_user` from an `initial` block (the one
  prerequisite whose failure has no fallback); an explicit clear checkbox for
  protected fields, since blank-means-keep had removed the only erase path;
  orphaned-session deletion when creation fails partway (the session already
  holds answers and nothing will ever resume it; #805 covers only downloaded
  packets); and a CSRF failure view so a cookies-blocked litigant gets the
  plain interview instead of a 403 dead end — that audience overlaps heavily
  with ours.

Known limitations, deliberate: `waiver_reasons` is never mapped (checkboxes
DADict, and the sensitive DV answer is better not stored cross-pathway — the
interview re-asks it). Cookie-blocked litigants always get the unprefilled
interview, since their answers live against a session they cannot hold.
`rel="noopener"` sits on a `<form>`, where browser support is patchier than
on anchors; the destination is our own docassemble. Interview packaging
(replacing manual Playground uploads) and a return link from the interview
are drafted as follow-up issues, not in this diff.

## Test plan

- [x] Manual, against the local bench (`make docassemble-up`, interviews +
  PDFs uploaded, `DOCASSEMBLE_*` set): answer name/county on the standard
  flow, click "Fill out your forms", land inside the interview with those
  questions skipped; finish and download the packet, values printed
  correctly; reuse the resume link from history, rejected (one-time); unset
  the API key, click again, plain interview opens (fallback); waiver track
  prefills name/county and still asks `waiver_reasons`.

---

### DIFF
--- a/docassemble/nd-name-change/README.md
+++ b/docassemble/nd-name-change/README.md
@@ -40,7 +40,7 @@ code block, not a bare `multi_user: True` key, which throws `DASourceError`
 Prereq: the bench is up. See [`docs/docassemble.md`](../../docs/docassemble.md).
 
 - Start the bench: `make docassemble-up`
-- Open `http://localhost:8100` and log in (fresh box default: `admin@example.com` / `password`)
+- Open `http://localhost:8100` and log in with the `DA_ADMIN_EMAIL` / `DA_ADMIN_PASSWORD` seeded on the bench's first boot (see [`docs/docassemble.md`](../../docs/docassemble.md))
 - Top-right menu → **Playground**
 - Upload the templates: in the Playground, open the **Templates** folder → upload
   `petition.pdf`, `declaration.pdf`, `notice.pdf`, `confidential-info.pdf`, and `order.pdf`
--- a/docassemble/nd-name-change/README.md
+++ b/docassemble/nd-name-change/README.md
@@ -26,6 +26,15 @@ forms (Petition, Declaration, Notice, Confidential, Order), the waiver track 4
 §12 (objections); they fork at entry, matching the Topic Flow corpus's
 standard/waiver split — no in-interview branching.
 
+## `multi_user = True` (prefill prereq)
+
+Both interviews open with an `initial` code block setting `multi_user = True`.
+The Litigant Portal prefill handoff creates a session through the API and
+hands the litigant a one-time resume link; without this flag the session is
+encrypted per-browser and that link cannot decrypt it. It has to be an `initial`
+code block, not a bare `multi_user: True` key, which throws `DASourceError`
+.
+
 ## Test it locally
 
 Prereq: the bench is up. See [`docs/docassemble.md`](../../docs/docassemble.md).
--- a/litigant_portal/app/services/docassemble.py
+++ b/litigant_portal/app/services/docassemble.py
@@ -10,7 +10,7 @@ decrypt it.
 """
 
 import logging
-from urllib.parse import parse_qs, urlparse
+from urllib.parse import parse_qs, urlparse, urlunparse
 
 import requests
 from django.conf import settings
@@ -46,7 +46,25 @@ def docassemble_session_create(*, interview_url: str, variables: dict) -> str:
 
     session = _session_new(api_root, api_key, interview)
     _variables_set(api_root, api_key, interview, session, variables)
-    return _resume_url(api_root, api_key, interview, session)
+    return _public(_resume_url(api_root, api_key, interview, session))
+
+
+def _public(resume_url: str) -> str:
+    """Swap in the litigant-facing origin, keeping the path and query.
+
+    docassemble builds the launch URL from the host we called it on, which on
+    a deployment is an internal address no browser can reach. Unset means the
+    URL comes back as docassemble built it.
+    """
+    public = settings.DOCASSEMBLE_PUBLIC_URL
+    if not public:
+        return resume_url
+    origin = urlparse(public)
+    return urlunparse(
+        urlparse(resume_url)._replace(
+            scheme=origin.scheme, netloc=origin.netloc
+        )
+    )
 
 
 def _target(interview_url: str) -> tuple[str, str]:
--- a/litigant_portal/app/services/docassemble.py
+++ b/litigant_portal/app/services/docassemble.py
@@ -1,0 +1,142 @@
+"""docassemble API client: start a prefilled interview session.
+
+Three calls, in order: create a session, set its variables, mint a one-time
+resume URL for the litigant's browser. Answers travel in a POST body, never
+in a URL, and the API key travels in a header.
+
+The target interview must set ``multi_user = True`` from an ``initial`` code
+block, or its session is encrypted per-browser and the resume URL cannot
+decrypt it.
+"""
+
+import logging
+from urllib.parse import parse_qs, urlparse
+
+import requests
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+_TIMEOUT = 10
+# Short enough that a leaked link in shared-terminal history is dead, long
+# enough for a slow redirect.
+_RESUME_EXPIRY_SECONDS = 600
+
+
+class DocassembleError(Exception):
+    """A session could not be created.
+
+    Callers fall back to the plain interview link: unprefilled beats a dead
+    button.
+    """
+
+
+def docassemble_session_create(*, interview_url: str, variables: dict) -> str:
+    """Return a one-time URL onto a new session prefilled with ``variables``.
+
+    ``interview_url`` is the plain launch link from the corpus packet section,
+    carrying the interview reference in its ``?i=`` parameter. Keys in
+    ``variables`` are interview-side names, which docassemble executes as
+    assignment statements, so they may only come from author-controlled YAML.
+    """
+    api_root, interview = _target(interview_url)
+    api_key = settings.DOCASSEMBLE_API_KEY
+    if not api_key:
+        raise DocassembleError("DOCASSEMBLE_API_KEY is unset")
+
+    session = _session_new(api_root, api_key, interview)
+    _variables_set(api_root, api_key, interview, session, variables)
+    return _resume_url(api_root, api_key, interview, session)
+
+
+def _target(interview_url: str) -> tuple[str, str]:
+    """``(api root, interview reference)`` for a launch link."""
+    parts = urlparse(interview_url)
+    references = parse_qs(parts.query).get("i", [])
+    if not references:
+        raise DocassembleError(f"No ?i= interview in {interview_url!r}")
+    # The API sits beside the launch route, so drop that last segment: QA's
+    # /interview/interview leaves the /interview/ path prefix in place.
+    root = settings.DOCASSEMBLE_BASE_URL or (
+        f"{parts.scheme}://{parts.netloc}{parts.path.rsplit('/', 1)[0]}"
+    )
+    return root.rstrip("/"), references[0]
+
+
+def _call(*, method: str, api_root: str, path: str, api_key: str, **kwargs):
+    url = f"{api_root}/api/{path}"
+    try:
+        response = requests.request(
+            method,
+            url,
+            headers={"X-API-Key": api_key},
+            timeout=_TIMEOUT,
+            **kwargs,
+        )
+        response.raise_for_status()
+    except requests.RequestException as exc:
+        # No variables in the message: it reaches logs and error reporting.
+        raise DocassembleError(f"docassemble {path} failed: {exc}") from exc
+    return response
+
+
+def _session_new(api_root: str, api_key: str, interview: str) -> str:
+    response = _call(
+        method="GET",
+        api_root=api_root,
+        path="session/new",
+        api_key=api_key,
+        params={"i": interview},
+    )
+    try:
+        return response.json()["session"]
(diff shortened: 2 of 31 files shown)


nicolearagao (commented):
@mitch-mitchel appreciate attentive eyes here, thank you

litigant_portal/app/services/docassemble.py
```diff
@@ -0,0 +1,185 @@
...
+
+def _target(interview_url: str) -> tuple[str, str]:
+    """``(api root, interview reference)`` for a launch link."""
+    parts = urlparse(interview_url)
+    references = parse_qs(parts.query).get("i", [])
+    if not references:
+        raise DocassembleError(f"No ?i= interview in {interview_url!r}")
+    # The API sits beside the launch route, so drop that last segment: QA's
+    # /interview/interview leaves the /interview/ path prefix in place.
+    root = settings.DOCASSEMBLE_BASE_URL or (
```
mitch-mitchel, line 103: Blocking. With `DOCASSEMBLE_BASE_URL` unset, the API root is whatever host the corpus names, and both corpora name `qa.litigantportal.com`. So an instance that sets `DOCASSEMBLE_API_KEY` and omits the base URL POSTs the key, plus the litigant's reviewed name, county and publication date, to QA.

FLP to FLP today, so misrouting rather than leakage. But it's the fail-open-on-missing-config shape you flagged on #873: drop one variable, get the wrong safe-looking behavior.

Fix without a schema change: raise `DocassembleError` from `_target()` when `DOCASSEMBLE_API_KEY` is set and `DOCASSEMBLE_BASE_URL` is not. That falls back to the plain interview link, which the view already handles.

litigant_portal/app/views/topic_flow.py
```diff
@@ -35,3 +54,72 @@ def topic_flow_download(request, court, topic, role, output_id):
...
+    corpus = registry.get(court, topic, role)
+    if corpus is None:
+        raise Http404(f"No Topic Flow for {court}/{topic}/{role}")
+
+    target = interview_target(corpus)
+    if target is None:
+        raise Http404(f"No interview handoff for {court}/{topic}/{role}")
+    interview_url, mapping = target
+
+    variables = prefill_variables(
```
mitch-mitchel, line 79: Non-blocking, a question. An empty payload still creates a session: three API calls and a `multi_user` (so unencrypted) session holding nothing, where `redirect(interview_url)` is the same experience.

`test_an_empty_variables_dict_still_creates_a_session` pins it as deliberate, so what does the session buy there? It's also the cookies-blocked litigant the CSRF fallback exists for, where no session means no answers means an empty payload every time.

litigant_portal/app/services/docassemble.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def _public(resume_url: str) -> str:
+    """Swap in the litigant-facing origin, keeping the path and query.
+
+    docassemble builds the launch URL from the host we called it on, which on
+    a deployment is an internal address no browser can reach. Unset means the
+    URL comes back as docassemble built it.
+    """
+    public = settings.DOCASSEMBLE_PUBLIC_URL
```
mitch-mitchel, line 84: Non-blocking. `_public()` keeps only scheme and netloc, so a `DOCASSEMBLE_PUBLIC_URL` carrying a path prefix is silently truncated. Defensible given the docstring says origin, but docassemble sits under `/interview/` on QA, which is where someone will set it wrong. A note in `.env.example` that it is origin-only would cover it.

docassemble/nd-name-change/README.md
```diff
@@ -26,12 +26,21 @@ forms (Petition, Declaration, Notice, Confidential, Order), the waiver track 4
...
+code block, not a bare `multi_user: True` key, which throws `DASourceError`
+.
+
 ## Test it locally
 
 Prereq: the bench is up. See [`docs/docassemble.md`](../../docs/docassemble.md).
 
 - Start the bench: `make docassemble-up`
-- Open `http://localhost:8100` and log in (fresh box default: `admin@example.com` / `password`)
+- Open `http://localhost:8100` and log in with the `DA_ADMIN_EMAIL` / `DA_ADMIN_PASSWORD` seeded on the bench's first boot (see [`docs/docassemble.md`](../../docs/docassemble.md))
```
mitch-mitchel, line 43: Non-blocking. `docs/docassemble.md` says these seed into an empty database only; this line doesn't. A pre-existing docassemble volume plus these vars gives a login failure with nothing pointing at `down -v`.

mitch-mitchel (changes requested):
Careful work. Three things worth naming: `_answer_map` guards `request.identity` behind a session check (the #871 bug, fixed here first); `test_the_nd_name_change_flows_are_actually_being_checked` keeps the drift guard from silently skipping itself, which is what makes `interview_prefill` trustworthy given a preset variable skips its own validation; and the handoff is a real form POST, so it degrades without JS.

One blocking ask, inline at `services/docassemble.py:103`. Everything else is non-blocking.

The larger version of that ask (launch URL and API root as validated per-environment config, and `docassemble.playground1:` not being a deployable reference) goes in its own issue rather than onto this PR.

litigant_portal/app/views/topic_flow.py
nicolearagao: Good question and this is something that my review skipped. The session buys nothing there, and like you said, the cookies-blocked litigant hits this path every time. Changed, when there is nothing to prefill the view now redirects straight to the plain interview link, no API calls, no empty session. The test that pinned the old behavior is gone, and the view tests that reached the client through an empty payload now store a reviewed answer first so they still cover what they were written for.

litigant_portal/app/services/docassemble.py
nicolearagao: Added a note in .env.example : origin only, a path prefix like /interview/ is silently dropped.

docassemble/nd-name-change/README.md
nicolearagao: Noted in the README: the seed only applies to an empty database, and a pre-existing volume keeps its old credentials, with a pointer at docker compose down -v.

litigant_portal/app/services/docassemble.py
nicolearagao: Fixed, the way you suggested: _target() now raises DocassembleError when DOCASSEMBLE_API_KEY is set and DOCASSEMBLE_BASE_URL is not, so the view falls back to the plain interview link instead of calling the host the corpus names. Added test_a_key_without_a_base_url_raises_without_calling_out, and the client tests now set the base URL explicitly since the fallback is gone. Also updated the .env.example comment to say the key needs the base URL beside it.

mitch-mitchel (approved):

mitch-mitchel (approved):
All four addressed, verified at f3dbd274.

The blocking one is properly closed: `_target()` now raises when `DOCASSEMBLE_API_KEY` is set without `DOCASSEMBLE_BASE_URL`, and the view's existing `DocassembleError` path turns that into the plain interview link rather than a 500. A misconfigured instance now degrades instead of leaking the key.

The tests are what sell it — `test_a_key_without_a_base_url_raises_without_calling_out` asserts `recorder.calls == []`, and the skip-path tests assert `docassemble.calls == []`. Asserting the absence of the request is the right shape for both of these; a test that only caught the exception would have passed while still calling out.

Nice catch on the CSRF fallback too — a cookie-blocked litigant reaching the plain link instead of a dead 403 is exactly the audience this handoff is for.

One thing left open, and it's yours to judge: claude's thread on `interview_target()` taking the first `PacketOutput` section with an `interview_url` when the schema permits several. Non-blocking, and a single-packet corpus can't hit it today.

litigant_portal/app/topic_flow/prefill.py
```diff
@@ -0,0 +1,33 @@
...
+    """``(launch url, {question id: interview variable})``, or None.
+
+    The first packet section carrying an interview_url wins; a corpus with no
+    interview handoff returns None.
+    """
+    for section in corpus.sections:
+        if isinstance(section, PacketOutput) and section.interview_url:
+            return section.interview_url, dict(section.interview_prefill)
+    return None
+
```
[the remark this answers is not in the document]
mitch-mitchel: Deferring this rather than changing it here. Checked the live corpora at f3dbd274: every file under `litigant_portal/content/` declares at most one `interview_url` (name-change standard and waiver have one each; both eviction flows have none), so first-match and correct-match are the same section today and no litigant can hit the ambiguity.

Worth revisiting when a corpus actually declares a second packet section with its own `interview_url` — at that point picking the first silently would be a real bug, and the fix is the handoff taking the section id rather than scanning. Not blocking this PR.

nicolearagao:
> All four addressed, verified at [f3dbd27](https://github.com/freelawproject/litigant-portal/commit/f3dbd27497e5c08a7cec17c30b2dcbd7dfa3d408).
> 
> The blocking one is properly closed: `_target()` now raises when `DOCASSEMBLE_API_KEY` is set without `DOCASSEMBLE_BASE_URL`, and the view's existing `DocassembleError` path turns that into the plain interview link rather than a 500. A misconfigured instance now degrades instead of leaking the key.
> 
> The tests are what sell it — `test_a_key_without_a_base_url_raises_without_calling_out` asserts `recorder.calls == []`, and the skip-path tests assert `docassemble.calls == []`. Asserting the absence of the request is the right shape for both of these; a test that only caught the exception would have passed while still calling out.
> 
> Nice catch on the CSRF fallback too — a cookie-blocked litigant reaching the plain link instead of a dead 403 is exactly the audience this handoff is for.
> 
> One thing left open, and it's yours to judge: claude's thread on `interview_target()` taking the first `PacketOutput` section with an `interview_url` when the schema permits several. Non-blocking, and a single-packet corpus can't hit it today.

Hey, I filed #889 to cover this. Merging... thanks for the review.

PR #871, opened 2026-09-09
Briefcase panel: show the litigant their case facts

### LINKED ISSUE #865: Briefcase panel: show the litigant their case facts
## What problem does this solve?

There's no briefcase. The right-hand aside renders `_agent_state.html` — a `<pre>` of `stateJson` headed "Briefcase body (live agent state for now)" — and it's gated to developers, so an anonymous litigant sees nothing at all.

Meanwhile the facts already exist. `VariableAnswer` stores them per identity, `variable_answer_list` reads them, and the guided flow writes them.

## What would you like to see?

Rebuild the panel from scratch as composed atoms and molecules, rendering real facts. Scope this first pass to what needs no migration, then reassess.

Grouping uses what the corpus already declares: `TopicFlowInterviewPage` gives an ordered `title` per group, and `TopicFlowInterviewVariable` places each variable on a page with its own order. Labels come from `Variable.label`, falling back to `name`.

### Assumptions

Stated so they can be argued with rather than discovered in review.

- **Selection rule.** Show answered variables only, grouped by the active flow's interview pages, with a single fallback group for answered variables that no page of that flow places. Not filtered by `required`, `is_global`, or `in_schema` — those become options once there's a real panel to look at.
- **The panel is always present**, with an empty state inside it. Org standard is no conditionally-hidden major layout sections, and it keeps the frame from shifting when the first fact lands.
- **Visible to everyone**, not developer-gated. The raw agent-state dump stays, moved into a developer-only disclosure inside the panel rather than being the panel.
- **Read-only.** Item-level edit and remove is #640.
- **Testable through the guided flow, not chat.** Nothing writes facts from chat yet — the agent's tools are `check_weather`, `load_topic_flow`, and `query_document`. A reviewer who opens a chat and sees an empty panel is seeing correct behavior.
- **No model change.** No section field, no order integer, nothing touching `Variable`. If the grouping proves wrong once it's visible, that is the reassessment, and the namespace half belongs to #751.

### Definition of done

- [ ] Panel renders on the right on desktop and in the existing mobile drawer, both fed by one component
- [ ] Facts grouped by interview page title in page order, variables in placement order, ungrouped bucket last
- [ ] Empty state renders when the identity has no answers
- [ ] `_agent_state.html` is no longer the panel body; raw state sits behind a developer disclosure
- [ ] Composed from existing atoms and molecules; any new component added to `style_guide.html` in the same commit
- [ ] View-level tests: grouping, ordering, empty state, anonymous visibility

## Relates

- #177 Briefcase unified session state model
- #751 case fact vocabulary — owns the namespace and the eventual typed grouping
- #640 item-level edit and remove
- #717 mobile drawer port
- #634 (closed) the responsive layout this reuses

mitch-mitchel:
The right-hand aside rendered a developer-gated JSON dump of agent state, so an anonymous litigant saw no briefcase at all while their facts sat unshown in `VariableAnswer`. The panel now renders those facts server-side on first paint, grouped for reading, composed from a new atom-level fact molecule and a panel organism.

Closes #865

## Grouping, and one deviation from the issue

The issue says group by "the active flow's interview pages." `chat_view` resolves no flow server-side, so grouping comes from each answered variable's own `TopicFlowInterviewVariable` placement instead. Same output when one flow is in play; when a variable sits on two flows' pages, the first in (topic, flow, page, placement) order wins, which keeps the grouping stable across requests.

No model change: `TopicFlowInterviewPage` already carries an ordered `title` and its docstring already calls the grouping cosmetic.

## Also here

- **The developer gate comes off** the desktop column, the mobile drawer, and the drawer trigger. The trigger was separately gated — leaving it would have made the drawer unopenable on mobile.
- **The raw agent-state dump survives** as an opt-in developer disclosure inside the panel. The opt-in exists because its Alpine bindings only resolve inside `x-data="chatApp"`, so the style guide would otherwise throw.
- **`VariableAnswer.display_value`** shapes jsonb for reading. Without it a multi-choice answer prints `['Unpaid rent']` at the litigant.
- Style guide gains both components, panel shown populated and empty.

## Test plan

20 tests in `test_briefcase.py` covering grouping order, placement order within a group, the trailing untitled group for unplaced answers, skipped empty pages, cleared and out-of-schema exclusion, cross-identity isolation, the chat-page context key, and `display_value`.

Four tests in `test_chat_page.py` were rewritten: they asserted the old contract, where anonymous users saw no briefcase chrome at all.

**Reviewing by hand:** open the guided flow and answer something, then open chat. Facts appear grouped by interview page. Opening a fresh chat shows the empty state — that is correct, not a bug. Nothing writes facts from chat yet; the agent's tools are `check_weather`, `load_topic_flow`, and `query_document`.

### DIFF
--- a/litigant_portal/app/selectors/topic_flow.py
+++ b/litigant_portal/app/selectors/topic_flow.py
@@ -52,20 +52,24 @@ def topic_flow_find(*, topic_slug: str, flow_slug: str) -> TopicFlow | None:
     )
 
 
-def variable_answer_list(*, identity) -> list[VariableAnswer]:
+def variable_answer_list(
+    *, identity, answered_only: bool = False
+) -> list[VariableAnswer]:
     """An identity's answers, ordered by variable name.
 
     Answers to variables the corpus no longer names (``in_schema=False``)
     are left out: sync keeps those rows so a migration can move them, but
     no form references them, so no surface should show them.
+
+    ``answered_only`` drops cleared answers (``value`` None), matching
+    ``variable_answer_map``: a cleared fact is not a fact.
     """
-    return list(
-        VariableAnswer.objects.filter(
-            identity=identity, variable__in_schema=True
-        )
-        .select_related("variable")
-        .order_by("variable__name")
+    answers = VariableAnswer.objects.filter(
+        identity=identity, variable__in_schema=True
     )
+    if answered_only:
+        answers = answers.filter(value__isnull=False)
+    return list(answers.select_related("variable").order_by("variable__name"))
 
 
 def variable_answer_map(*, identity, names: list[str]) -> dict:
@@ -86,7 +90,7 @@ def variable_answer_map(*, identity, names: list[str]) -> dict:
     )
 
 
-def briefcase_groups(*, identity) -> list[dict]:
+def variable_answer_groups(*, identity) -> list[dict]:
     """An identity's answers, grouped for reading by interview page.
 
     Grouping comes from each answered variable's own placement
@@ -104,27 +108,25 @@ def briefcase_groups(*, identity) -> list[dict]:
     matching ``variable_answer_map``: a cleared fact is not a fact, and no
     surface should show a variable the corpus no longer names.
     """
-    answers = list(
-        VariableAnswer.objects.filter(
-            identity=identity, variable__in_schema=True, value__isnull=False
-        )
-        .select_related("variable")
-        .order_by("variable__name")
-    )
+    answers = variable_answer_list(identity=identity, answered_only=True)
     if not answers:
         return []
 
     answer_by_variable = {a.variable_id: a for a in answers}
     placements = (
         TopicFlowInterviewVariable.objects.filter(
-            variable_id__in=answer_by_variable
+            variable_id__in=answer_by_variable, page__flow__enabled=True
         )
         .select_related("page")
         .order_by(
             "page__flow__topic__order",
+            "page__flow__topic__created_at",
             "page__flow__order",
+            "page__flow__created_at",
             "page__order",
+            "page__created_at",
             "order",
+            "created_at",
         )
     )
 
--- a/litigant_portal/app/selectors/topic_flow.py
+++ b/litigant_portal/app/selectors/topic_flow.py
@@ -1,7 +1,12 @@
 from django.core.cache import cache
 
 from litigant_portal.app.cache import TOPIC_LIST_CACHE_KEY
-from litigant_portal.app.models import Topic, TopicFlow, VariableAnswer
+from litigant_portal.app.models import (
+    Topic,
+    TopicFlow,
+    TopicFlowInterviewVariable,
+    VariableAnswer,
+)
 
 
 def topic_list() -> list[Topic]:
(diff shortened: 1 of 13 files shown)


litigant_portal/app/views/pages.py
```diff
@@ -38,8 +45,18 @@ def home(request):
...
+
+    The briefcase renders server-side from the visitor's stored facts, so the
+    panel is populated on first paint rather than waiting on a fetch. The key
+    is always present, empty list included — the panel is a fixed part of the
+    frame and its empty state is a render, not an absence.
+    """
+    return render(
+        request,
+        "pages/chat/index.html",
+        {"briefcase_groups": briefcase_groups(identity=request.identity)},
```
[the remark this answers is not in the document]
nicolearagao: yes, needs to be fixed.

litigant_portal/app/tests/test_briefcase.py
```diff
@@ -0,0 +1,219 @@
...
+        # The panel is always rendered, so the context key must always exist:
+        # a missing key and an empty list are different bugs in the template.
+        response = self.client.get(reverse("pages:chat"))
+
+        self.assertEqual(response.context["briefcase_groups"], [])
+
+
+class DisplayValueTests(TestCase):
+    """``VariableAnswer.display_value`` shapes jsonb for reading. No DB —
+    the property only touches the in-memory value."""
```
nicolearagao, line 192: says "No DB" but uses TestCase, which always opens a database connection. So it would fail under tox -e fast, where there is no Postgres. Use SimpleTestCase instead.

litigant_portal/app/selectors/topic_flow.py
```diff
@@ -79,3 +84,67 @@ def variable_answer_map(*, identity, names: list[str]) -> dict:
...
+        .select_related("variable")
+        .order_by("variable__name")
+    )
+    if not answers:
+        return []
+
+    answer_by_variable = {a.variable_id: a for a in answers}
+    placements = (
+        TopicFlowInterviewVariable.objects.filter(
+            variable_id__in=answer_by_variable
```
nicolearagao, line 120: The placements query in briefcase_groups doesn't filter on page__flow__enabled=True, but every other flow selector in the file does. enabled defaults to False, so a draft flow's page title can show up as a heading in a real user's briefcase.

litigant_portal/app/views/pages.py
```diff
@@ -174,7 +191,48 @@ def accessibility(request):
...
+    """Unsaved sample facts for the style guide's briefcase entry.
+
+    Built in memory rather than queried so the page renders the same on a
+    fresh database, and so browsing the style guide never shows a real
+    visitor's answers.
+    """
+
+    def fact(name, label, value):
+        return VariableAnswer(
+            variable=Variable(name=name, label=label), value=value
```
nicolearagao, line 211: The fact() helper builds Variable(name=name, label=label) with no data_type, so it defaults to TEXT. The fix is letting fact() accept a data type and passing it for this one entry, e.g.:
```
def fact(name, label, value, data_type=VariableDataType.TEXT):
    return VariableAnswer(
        variable=Variable(name=name, label=label, data_type=data_type), value=value
    )
...
fact("received_date", "Date received", "2026-09-01", VariableDataType.DATE),
```

litigant_portal/app/selectors/topic_flow.py
```diff
@@ -79,3 +84,67 @@ def variable_answer_map(*, identity, names: list[str]) -> dict:
...
+        .order_by(
+            "page__flow__topic__order",
+            "page__flow__order",
+            "page__order",
+            "order",
+        )
+    )
+
+    # Walking placements in that order builds both the page sequence and the
+    # sequence within each page, so neither needs a second sort.
```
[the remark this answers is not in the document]
nicolearagao: yes, tiebreaker here would be nice

litigant_portal/app/selectors/topic_flow.py
```diff
@@ -79,3 +84,67 @@ def variable_answer_map(*, identity, names: list[str]) -> dict:
...
+
+    Pages that place nothing this identity answered are skipped rather than
+    rendered as empty headings, and answers with no placement collect in a
+    final untitled group.
+
+    Cleared answers (``value`` None) and out-of-schema variables are left out,
+    matching ``variable_answer_map``: a cleared fact is not a fact, and no
+    surface should show a variable the corpus no longer names.
+    """
+    answers = list(
```
nicolearagao, line 107: briefcase_groups rebuilds almost the same query as variable_answer_list, only adding value__isnull=False. If the two stay separate, a future change to which answers are visible could update one and miss the other. variable_answer_list has no production callers right now, so it's free to reuse or adapt.

litigant_portal/app/selectors/topic_flow.py
```diff
@@ -79,3 +84,67 @@ def variable_answer_map(*, identity, names: list[str]) -> dict:
             value__isnull=False,
         ).values_list("variable__name", "value")
     )
+
+
+def briefcase_groups(*, identity) -> list[dict]:
```
nicolearagao, line 89: CLAUDE.md says selectors are named {model_name}_{action}, not after the page that uses them. This one reads VariableAnswer rows, so variable_answer_groups would fit that pattern better.

nicolearagao (changes requested):
Nice, thanks for also updating the style guide. A few things to fix before merge.

mitch-mitchel:
@nicolearagao changes are in

nicolearagao (approved):
Verified all the fixes were addressed. Thanks Mitch, LGTM

Issue #861, opened 2026-09-08
Fail loudly on missing or weak admin credentials and SECRET_KEY


mitch-mitchel:
### What needs to change

Three places where a missing or weak credential fails open instead of failing loudly. All app-side, independent of the infra move (#859).

**1. `bootstrap_superuser` bypasses password validation.** `create_superuser()` sets the password directly and `AUTH_PASSWORD_VALIDATORS` only run in forms, so `SUPERUSER_PASSWORD=admin` is accepted silently and produces a working superuser. Call `validate_password()` before creating and fail with a non-zero exit.

**2. `SECRET_KEY` fails open.** `settings.py:32` is `os.environ.get("SECRET_KEY") or get_random_secret_key()`. With the key missing in a deployed environment, each gunicorn worker generates its own (three workers, no `--preload`), so sessions and CSRF break nondeterministically instead of the deploy failing. Raise `ImproperlyConfigured` when `not DEBUG`.

**3. `/django-admin/login/` has no brute-force protection.** django-ratelimit is applied across `views/admin.py` and `views/assistant.py`, and allauth 65.14.1 throttles its own login via default `ACCOUNT_RATE_LIMITS` — but Django's admin login isn't ours to decorate, and `SitePasswordMiddleware` takes unlimited guesses against a plain `==` comparison. Both want a rate limit.

Also small and in scope: `docker-compose.docassemble.yml`'s header comment says the docassemble default login is `admin@admin.com`. The current image uses `admin@example.com` (`create_tables.py:178`); `admin@admin.com` appears nowhere in `jhpyle/docassemble`.

### Why

Four court-partner instances land under #859, each standing up its own Django superuser. A silent fallback multiplies by five. Found while checking whether the admin surfaces carry default credentials — they don't, but nothing stops a weak one.

### Definition of done

- [ ] `bootstrap_superuser` validates the password and exits non-zero on a weak one
- [ ] A missing `SECRET_KEY` raises `ImproperlyConfigured` when `DEBUG` is off
- [ ] Admin login and the site-password gate are rate-limited
- [ ] The docassemble default-login comment names the right email

PR #873, opened 2026-09-09
Fail startup when a deployed environment has no site password

### LINKED ISSUE #870: SITE_PASSWORD fails open: blank gate serves the site with no signal
### What needs to change

`SITE_PASSWORD` fails open. `settings.py` reads it as `os.environ.get("SITE_PASSWORD", "")` and `SitePasswordMiddleware` gates on `not password`, so a blank value disables the wall entirely and serves every page to anyone. That is the correct behavior in dev. In a deployed environment it means a typo, an unset secret, or a dropped env var takes the gate down with no signal beyond a compose warning scrolling past at boot:

```
WARN[0000] The "SITE_PASSWORD" variable is not set. Defaulting to a blank string.
```

Two changes:

**1. A system check that refuses a blank gate outside dev.** Register a check that raises an `Error` when `DEPLOYMENT_ENV` is `qa` or `prod` and `SITE_PASSWORD` is empty. Checks already run at startup and on every management command (`checks/prompts.py`, `checks/corpus.py`), so a failure stops the deploy rather than serving an open site.

Going public is then a deliberate, reviewable act: add the check's id to `SILENCED_SYSTEM_CHECKS`. That keeps the policy out of the code path — the wall comes down by an explicit config change with a name attached, not by a value quietly going missing.

**2. Declare the dev default in compose.** `docker-compose.yml` passes `${SITE_PASSWORD}`; make it `${SITE_PASSWORD:-}` so a blank dev gate reads as intentional instead of warning every boot. Silencing the warning only makes sense alongside change 1 — otherwise it removes the one signal there is.

### Why

The gate is the only thing standing between the public and a pre-launch portal, and its failure mode is invisible. Note the asymmetry with the rest of the config: `ALLOWED_HOSTS` deliberately fails closed when unset, and `DEPLOYMENT_ENV` warns on an unrecognized value. This is the security control that does neither.

### Definition of done

- [ ] System check errors when `DEPLOYMENT_ENV` is qa or prod and `SITE_PASSWORD` is blank
- [ ] The check's id is documented, along with how to silence it deliberately
- [ ] `docker-compose.yml` declares the empty default explicitly
- [ ] Tests cover: blank in prod errors, blank in dev passes, set in prod passes, silenced check passes

### Relates

Sibling of #861 (fail loudly on missing or weak admin credentials and `SECRET_KEY`), which covers the same failure mode for `SECRET_KEY`, `bootstrap_superuser`, and rate limiting. Different blast radius: this one exposes the whole site rather than one account.

mitch-mitchel:
`SitePasswordMiddleware` gates on `not password`, so a blank `SITE_PASSWORD` disables the site-wide wall and serves every page. That is the right default in local dev and a silent failure anywhere else: a dropped secret or a typo takes the gate down, with nothing louder than a compose warning scrolling past at boot.

```
WARN[0000] The "SITE_PASSWORD" variable is not set. Defaulting to a blank string.
```

Closes #870, sub-issue of #861.

## What changes

A system check (`litigant_portal.E001`) errors at startup when the environment is deployed and the password is blank or whitespace-only. Checks already run on startup and every management command, so this stops a deploy rather than serving an open site.

Whitespace-only is its own failure, not a variant of blank: `not password` reads `"   "` as a *live* gate, so every visitor is locked out by a value nobody can type.

Compose now declares the empty dev default explicitly (`${SITE_PASSWORD:-}`). That only makes sense alongside the check — on its own it would remove the one signal there was.

## One deviation from the issue

#870 says the check should fire for `qa` or `prod`. It fires for anything that isn't `dev`. `settings.py` keeps an unrecognized `DEPLOYMENT_ENV` as-is and only warns, so matching the two known labels would wave `DEPLOYMENT_ENV=staging` through with the wall down. Failing closed matches how `ALLOWED_HOSTS` already behaves.

## Taking the wall down on purpose

```python
SILENCED_SYSTEM_CHECKS = ["litigant_portal.E001"]
```

Going public becomes an explicit config change with a name attached, rather than a secret quietly going missing. Documented in `docs/wiki/SECURITY.md`.

## Test plan

Seven tests in `test_site_password_check.py`: blank in prod and qa error, blank in dev passes, a set password in prod passes, whitespace-only errors, unset errors, and an unrecognized environment is treated as deployed.

**Before merging, worth confirming:** whether QA and prod actually have `SITE_PASSWORD` set. That question is still open pending cluster access. If either is running blank today, this turns an open site into a failed boot — which is the intent, but better to know now than during a deploy.

### DIFF
--- a/litigant_portal/app/checks/site_password.py
+++ b/litigant_portal/app/checks/site_password.py
@@ -1,0 +1,45 @@
+"""Django system check for the temporary site-wide password gate.
+
+``SitePasswordMiddleware`` disables itself when ``SITE_PASSWORD`` is empty,
+which is the right default for local development and a silent failure anywhere
+else: a dropped secret or a typo takes the wall down and serves a pre-launch
+portal to the public, with nothing louder than a compose warning at boot.
+
+This check turns that into a startup failure. Taking the wall down on purpose
+means silencing the check by id, which is a reviewable config change with a
+name attached rather than a value quietly going missing.
+"""
+
+from django.conf import settings
+from django.core.checks import Error, Tags, register
+
+SITE_PASSWORD_CHECK_ID = "litigant_portal.E001"
+
+# Anything that isn't local development counts as deployed. settings.py keeps
+# an unrecognized DEPLOYMENT_ENV as-is and only warns, so testing for "not dev"
+# fails closed where testing for "qa or prod" would wave a typo through.
+_LOCAL_ENV = "dev"
+
+
+@register(Tags.security, deploy=False)
+def check_site_password_set_when_deployed(app_configs, **kwargs):
+    """Error when a deployed environment has no usable site password."""
+    if getattr(settings, "DEPLOYMENT_ENV", _LOCAL_ENV) == _LOCAL_ENV:
+        return []
+
+    password = getattr(settings, "SITE_PASSWORD", "") or ""
+    if password.strip():
+        return []
+
+    return [
+        Error(
+            "SITE_PASSWORD is blank, so the site-wide password gate is off "
+            "and every page is public.",
+            hint=(
+                "Set SITE_PASSWORD in this environment. If the portal is "
+                "meant to be public, take the gate down deliberately by "
+                f"adding '{SITE_PASSWORD_CHECK_ID}' to SILENCED_SYSTEM_CHECKS."
+            ),
+            id=SITE_PASSWORD_CHECK_ID,
+        )
+    ]
--- a/litigant_portal/app/tests/test_site_password_check.py
+++ b/litigant_portal/app/tests/test_site_password_check.py
@@ -1,0 +1,57 @@
+"""The deployed-environment guard on the site-password gate.
+
+`SitePasswordMiddleware` treats a blank `SITE_PASSWORD` as "no gate", which is
+right in dev and silently serves a pre-launch portal to the public anywhere
+else. The check turns that into a failed startup.
+"""
+
+from django.core.checks import Error
+from django.test import TestCase, override_settings
+
+from litigant_portal.app.checks.site_password import (
+    SITE_PASSWORD_CHECK_ID,
+    check_site_password_set_when_deployed,
+)
+
+
+class SitePasswordCheckTests(TestCase):
+    def _run(self):
+        return check_site_password_set_when_deployed(app_configs=None)
+
+    @override_settings(DEPLOYMENT_ENV="prod", SITE_PASSWORD="")
+    def test_blank_gate_in_prod_is_an_error(self):
+        errors = self._run()
+        self.assertEqual(len(errors), 1)
+        self.assertIsInstance(errors[0], Error)
+        self.assertEqual(errors[0].id, SITE_PASSWORD_CHECK_ID)
+
+    @override_settings(DEPLOYMENT_ENV="qa", SITE_PASSWORD="")
+    def test_blank_gate_in_qa_is_an_error(self):
+        self.assertEqual(len(self._run()), 1)
+
+    @override_settings(DEPLOYMENT_ENV="dev", SITE_PASSWORD="")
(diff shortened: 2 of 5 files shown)


litigant_portal/app/tests/test_site_password_check.py
```diff
@@ -0,0 +1,57 @@
...
+from django.core.checks import Error
+from django.test import TestCase, override_settings
+
+from litigant_portal.app.checks.site_password import (
+    SITE_PASSWORD_CHECK_ID,
+    check_site_password_set_when_deployed,
+)
+
+
+class SitePasswordCheckTests(TestCase):
```
nicolearagao, line 17: Same case here from the other PR. The check function touches no database, but django.test.TestCase always creates a test DB, and the file has no @pytest.mark.postgres marker, so tox -e fast collects it and dies. Changing to SimpleTestCase would fix this.

litigant_portal/app/checks/site_password.py
```diff
@@ -0,0 +1,45 @@
...
+# Anything that isn't local development counts as deployed. settings.py keeps
+# an unrecognized DEPLOYMENT_ENV as-is and only warns, so testing for "not dev"
+# fails closed where testing for "qa or prod" would wave a typo through.
+_LOCAL_ENV = "dev"
+
+
+@register(Tags.security, deploy=False)
+def check_site_password_set_when_deployed(app_configs, **kwargs):
+    """Error when a deployed environment has no usable site password."""
+    if getattr(settings, "DEPLOYMENT_ENV", _LOCAL_ENV) == _LOCAL_ENV:
```
nicolearagao, line 27: DEPLOYMENT_ENV defaults to "dev" when unset (settings.py:18), so a deploy that drops both it and SITE_PASSWORD passes the check with the wall down. Same fail-closed argument you made for staging. Adding DEBUG is False as a second deployed-signal would cover it.

litigant_portal/app/tests/test_site_password_check.py
```diff
@@ -0,0 +1,57 @@
...
+    @override_settings(DEPLOYMENT_ENV="qa", SITE_PASSWORD="")
+    def test_blank_gate_in_qa_is_an_error(self):
+        self.assertEqual(len(self._run()), 1)
+
+    @override_settings(DEPLOYMENT_ENV="dev", SITE_PASSWORD="")
+    def test_blank_gate_in_dev_is_fine(self):
+        # The dev default. Erroring here would make every fresh clone fail to
+        # start, which is how a check gets silenced permanently.
+        self.assertEqual(self._run(), [])
+
```
nicolearagao, line 37: All 7 tests call the function directly, so nothing exercises registration or the SILENCED_SYSTEM_CHECKS path that SECURITY.md documents as the sanctioned way to go public. A call_command("check") test under override_settings(..., SILENCED_SYSTEM_CHECKS=[SITE_PASSWORD_CHECK_ID]) would pin both.

nicolearagao (changes requested):

mitch-mitchel:
Closing: this check defends `SitePasswordMiddleware`, and the middleware is being replaced rather than hardened. See #885.

Three holes the check cannot see. `/api/` is exempt (`middleware.py:22-24`) and the chat stream is `/api/agents/assistant/stream/`, so the one endpoint that costs money per request is open. `/interview/*` never reaches Django, which is how the docassemble Playground sat open behind the wall. And the README mounts add-on services by path, so every future one is outside a Django middleware by construction. A check catching a blank `SITE_PASSWORD` guards the smaller half.

@nicolearagao your `DEPLOYMENT_ENV` finding stands and would have been the fix if we kept this: a deploy dropping both variables sails through with the wall down. Same for `SimpleTestCase` and the missing `SILENCED_SYSTEM_CHECKS` coverage. Thanks for the read.

The middleware stays in place until the ingress gate lands. It just is not a gate.

PR #874, opened 2026-09-09
[867] Add QA deploy workflow to EKS qa-litigant

chaco-fl:


mitch-mitchel:
thanks @chaco-fl 
2 requests
- can we have the tests -> build -> deploy flow that prod follows? we won't deploy broken branches to QA, and should catch if they are
- as per slack convo, branch deploy in here too ~~i would prefer a manual Github action vs merge to a branch name, similar to the [Digital Ocean one](https://github.com/freelawproject/litigant-portal/blob/main/.github/workflows/qa-deploy-do.yml). is that possible to setup?~~

mitch-mitchel:
Test gate looks right, thanks. One thing will stop it running: `tests.yml` declares `workflow_call:` with no inputs, so passing `with: ref:` is a validation error and the run fails before any job starts. Prod calls it bare, which is why it works there.

Deleting `with:` would pass validation but test the wrong commit — a reusable workflow checks out the caller's SHA, so a manual dispatch of some branch would run tests against `main` while `build` builds the dispatched ref.

So `tests.yml` needs the input:

```yaml
on:
  workflow_call:
    inputs:
      ref:
        type: string
        required: false
```

and its checkout using `ref: ${{ inputs.ref || github.sha }}`. Prod's bare call keeps working, and QA tests what it deploys.

Small one while you're in there: `deploy_ref` defaults to `qa`, which will be a dead default if that branch ever goes away — `main` is probably the safer default.

ERosendo:
Hi @mitch-mitchel! I think your comment might be looking at an older state of the branch, so I wanted to double-check that we're looking at the same thing 🤔 

The `ref` input in `tests.yml` was already added in [64e8f26](https://github.com/freelawproject/litigant-portal/pull/874/commits/64e8f26e4c69d563ea6e7d23be0dfd42d18221b8), which is also the same commit that added the `tests:` job to `qa-deploy.yml`

Both pieces landed together, so I don't think there was a commit on this PR where the with: ref: call existed without the corresponding input (the first commit didn't have the gate)

Am I missing something here? Happy to take another look if you're seeing a scenario where this could still fail.

mitch-mitchel:
@ERosendo nope, my bad, you're right, this is why it's better to do code reviews in the morning.

mitch-mitchel (approved):
LGTM!

nicolearagao (approved):
LGTM

PR #876, opened 2026-09-10
feat(agent): direct runtime and test page

Sinjhin:
## What changed

Implements one scoped Bedrock response per submission through `PortalAgent`, with async run handles and a development page that displays streaming text, status, and outcome events.

Base branch: `860-01-agent-package-and-contract`.

- Add Direct execution and a synchronous newline-delimited JSON (NDJSON) iterator for Django responses. Enforce the active-time limit and close owned tasks, model streams, and HTTP clients on completion, failure, cancellation, or explicit closure.
- Keep model behavior in generic `lp_agent/flows/`, execution in `runtimes/`, and provider/store construction in `adapters/`. Temporary memory stores retain run state, model output, and instruction snapshots; reasoning metadata stays outside public events.
- Wire `PortalAgent` to validate the explicit saved user identity and supply server configuration. `AgentIdentity` and `ResourceScope` hold the shared and scoped services. Court/topic choices belong to the host, and an omitted judge reuses the primary model client.
- Add `/dev/agent/` with court, topic, model, and active-time controls. Access requires login, developer permission, and `LP_AGENT_DEV_ENABLED`; local Docker Compose enables the flag, while other environments default to disabled. Each Send starts an independent response.
- Add typed database, S3, and vector corpus interfaces that raise `NotImplementedError`, plus temporary file retrieval using the supplied resource root. The file reader returns YAML content with source paths and rejects missing, unreadable, or escaping paths. Flows can call these functions by court/topic keys; retrieval is not connected to model context yet.

This branch uses instance-local storage. Durable persistence, conversation continuation/recovery, scope discovery, tools, queue/steer, and Workers execution remain unimplemented. Main-chat integration is separate work.

Refs #860.

## Test plan

- [x] Pre-commit hooks passed for the changed files; `make test` passed **986 tests and 27 subtests**.
- [x] `tox -e agent -- -q` passed **216 tests** in the isolated core environment, without Django or provider SDKs installed.
- [x] Regression tests cover identity/scope authorization, initialization callbacks, judge defaults, checkpoint ownership, timeout/cancellation/shutdown, and file-corpus boundaries.
- [x] Provider tests exercise native and translated LiteLLM streams with controlled responses, including metadata preservation, incomplete output, safe failures, and connection cleanup. Django tests cover access gates, CSRF, invalid inputs, incremental output, and response closure.
- [x] Browser checks against the rendered page with controlled HTTP responses passed for topic filtering, incremental text, control state, plain-text rendering, error recovery, and desktop/mobile layout, with no JavaScript errors.

Live Bedrock verification through an authenticated browser session remains pending.

---


litigant_portal/app/views/agent.py
```diff
@@ -0,0 +1,123 @@
...
+            "invalid_choice": "Select a topic available for this court."
+        }
+    )
+    model = forms.ChoiceField(
+        choices=MODEL_CHOICES,
+        error_messages={
+            "invalid_choice": "Select an available Bedrock model."
+        },
+    )
+    max_active_seconds = forms.FloatField(min_value=0.1)
```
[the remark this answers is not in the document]
Sinjhin: Deferred until later PRs where we implement other guards.

litigant_portal/app/static/js/agent_development.js
```diff
@@ -0,0 +1,152 @@
+const agentDevelopmentForm = document.getElementById('agent-development-form')
```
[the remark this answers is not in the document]
Sinjhin: Resolved with: [ae2a64b](https://github.com/freelawproject/litigant-portal/pull/876/commits/ae2a64b075d9fee64faa47eee570de592ae74dde)

litigant_portal/app/views/agent.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def clean(self):
+        data = super().clean()
+        if not data.get("message", "").strip():
+            self.add_error("message", "Enter a message.")
+        return data
+
+
+@login_required
+@permission_required("app.manage_developers", raise_exception=True)
```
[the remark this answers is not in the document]
Sinjhin: Resolved with: [ae2a64b](https://github.com/freelawproject/litigant-portal/pull/876/commits/ae2a64b075d9fee64faa47eee570de592ae74dde)

litigant_portal/app/static/js/agent_development.js
```diff
@@ -0,0 +1,147 @@
...
+    selectCourt() {
+      const form = this.$root
+      this.selectedCourt = form.elements.court.value
+      const topic = form.elements.topic
+      const selected = topic.selectedOptions[0]
+      if (selected?.value && selected.dataset.court !== this.selectedCourt)
+        topic.value = ''
+    },
+
+    async send() {
```
nicolearagao, line 40: found while manually testing. send() reads the textarea into FormData but never clears it, and the finally block re-focuses the textarea with the stale text, so a second Enter re-submits the old message. Fix should be something like after FormData capture: form.elements.message.value = ''

lp_agent/flows/engagement.py
```diff
@@ -0,0 +1,269 @@
...
+    ) -> RunOutcome:
+        try:
+            await self._save("running", emit)
+            if cancelled:
+                raise asyncio.CancelledError
+            async with asyncio.timeout(self.limits.max_active_seconds):
+                outcome = await self._model_response(emit)
+        except asyncio.CancelledError:
+            outcome = CancelledOutcome(**self.reference)
+        except TimeoutError:
```
nicolearagao: except TimeoutError catches more than the active-time limit. A provider connect timeout is also a TimeoutError (since Python 3.11), and the _save("running") call at line 151 is inside the try but outside the timeout scope. Both would be reported as active_time_limit, which points debugging at the wrong place.

Suggested fix: async with asyncio.timeout(...) as tm and treat it as the time limit only when tm.expired() is true.

lp_agent/flows/prompts.py
```diff
@@ -0,0 +1,18 @@
...
+
+from lp_agent.types import Scope
+
+
+def system_prompt(scope: Scope) -> str:
+    """
+    Describe the selected context and the current lack of retrieval tools.
+    """
+    return (
+        "You are the Litigant Portal assistant. Respond clearly and briefly. "
```
nicolearagao, line 17: The prompt has good honesty rules (no invented facts, no claimed searches), but none of the project content rules: no UPL boundary, no plain-language guidance for the audience. Default model behavior might break all of these.

If this is a placeholder, can the PR say where the real prompt content lands? If not, these rules should be added here.

@mitch-mitchel pretty sure I am missing something, pls chime in

lp_agent/flows/engagement.py
```diff
@@ -0,0 +1,269 @@
...
+            )
+        return outcome
+
+    async def finish(
+        self, outcome: RunOutcome, emit: Callable[[EventPayload], None]
+    ) -> None:
+        """
+        Commit the terminal state before publishing its outcome.
+        """
+        await self._save(outcome.state, emit, outcome)
```
nicolearagao, line 195: finish() has no exception handling, so a failing terminal commit_checkpoint (whose payload includes the full prompt) raises a raw storage exception to the host, bypassing the sanitization used on the model path.

litigant_portal/app/views/agent.py
```diff
@@ -0,0 +1,124 @@
...
+    try:
+        agent = PortalAgent(
+            identity=request.identity,
+            court=data["court"],
+            topic=data["topic"],
+            model=data["model"],
+            runtime="Direct",
+            limits=RunLimits(max_active_seconds=data["max_active_seconds"]),
+        )
+    except (AgentValidationError, ValidationError):
```
nicolearagao, line 126: This except also catches "A Bedrock API key is required" from BedrockClient when AWS_BEARER_TOKEN_BEDROCK is unset, so a misconfigured environment shows as a client 400 "Invalid agent configuration." Suggest logging the exception before returning 400.

lp_agent/adapters/bedrock.py
```diff
@@ -0,0 +1,258 @@
...
+    ModelMessage,
+    ModelOutputItem,
+    ModelRequest,
+    ModelTextDelta,
+    OutputText,
+    ReasoningItem,
+    ToolCall,
+)
+
+MODEL_CHOICES = (
```
nicolearagao, line 29: MODEL_CHOICES and the app's BedrockModel enum (litigant_portal/app/models/choices.py) are two hand-maintained lists that already diverge (Haiku is deliberately absent here). The fallback in views/agent.py:36-38 handles it today, but nothing catches silent drift when a model is added to one list only. A small cross-list assertion test would pin the relationship

nicolearagao (changes requested):
Nice!

Test-coverage gaps worth a follow-up (none blocking):

- RunHandle.cancel() is never called directly in any test. Cancellation is only exercised through agent.aclose(), so the public cancel-mid-stream path (observer ends via the sentinel, result() returns CancelledOutcome) is uncovered.
Engagement.respond() always raises "This run has no pending question.", but no test pins that contract.
- Attaching a second events() observer should be rejected (runtimes/direct.py:104), but there is no test for it.
- In adapters/bedrock.py, output items are validated with extra="forbid", so a single unknown metadata field from the provider fails the whole stream and on the error path can also drop the already-buffered good items. There is no test asserting what should happen in that case, and it's probably the adapter's most likely production failure mode, so the intended behavior should be a tested decision.

lp_agent/flows/prompts.py
Sinjhin: Just a placeholder, yeah.

litigant_portal/app/static/js/agent_development.js
Sinjhin: Resolved with: [5db940b](https://github.com/freelawproject/litigant-portal/pull/876/commits/5db940bfbd6a6fe5b2ad50a7d2c2a7f04717d9b0)

lp_agent/flows/engagement.py
Sinjhin: Resolved with: [5db940b](https://github.com/freelawproject/litigant-portal/pull/876/commits/5db940bfbd6a6fe5b2ad50a7d2c2a7f04717d9b0)

lp_agent/flows/prompts.py
Sinjhin: Resolved with: [5db940b](https://github.com/freelawproject/litigant-portal/pull/876/commits/5db940bfbd6a6fe5b2ad50a7d2c2a7f04717d9b0)

(I just put in README and file it's a placeholder for now)

lp_agent/flows/engagement.py
Sinjhin: Resolved with: [5db940b](https://github.com/freelawproject/litigant-portal/pull/876/commits/5db940bfbd6a6fe5b2ad50a7d2c2a7f04717d9b0)

Both initial and final checkpoint failures now surface a safe storage error. Completion is reported only after the final state has been saved.

litigant_portal/app/views/agent.py
Sinjhin: Resolved with: [5db940b](https://github.com/freelawproject/litigant-portal/pull/876/commits/5db940bfbd6a6fe5b2ad50a7d2c2a7f04717d9b0)

A missing server key now returns 503 with a safe log message; invalid submitted input still returns 400.

lp_agent/adapters/bedrock.py
Sinjhin: Resolved with: [5db940b](https://github.com/freelawproject/litigant-portal/pull/876/commits/5db940bfbd6a6fe5b2ad50a7d2c2a7f04717d9b0)

nicolearagao (approved):
LGTM

PR #877, opened 2026-09-10
feat(agent): 179 01 agent evaluation suite

Sinjhin:
## What changed

Adds a terminal evaluation suite for comparing answer quality, response time, and estimated API cost across a raw Bedrock model, the existing `LitigantAssistant` chat engine, and `PortalAgent` using its Direct runtime.

Base branch: `179-01-agent-evaluation-baseline`.

- Add `make agent-eval` with configurable systems, models, cases, repetitions, judge, scoring weights, and pricing. The default matrix is three systems × two models (Luna/Terra) × 14 cases × three repetitions: **252 answers**, followed by separate Sol judging. Raw receives only the user question.
- Keep the harness and its independent uv environment in `scripts/agent_eval/`. Orchestration, raw calls, judging, and charts run locally; old/new agents execute through a temporary worker in the existing Django development container. The only changes outside this directory are the Make target and ignoring `evals/`.
- Add eight North Dakota reference cases covering procedure, fees, uncertainty, legal strategy, and escalation, plus six fictional chicken-law cases. Paired corpus variants change fees and procedural order, with eligibility boundary cases to measure whether answers follow the supplied content. Freeze references, configuration, and source provenance with each run.
- Score seven dimensions with a default 70% factual / 30% qualitative weighting. Validate judge evidence and expected values, record critical failures, and distinguish answering from deferral. Charts retain weighted quality and mark critical failures bright red; JSON also records a zeroed score for each answer with a critical failure.
- Save answers, traces, judgments, timing, token usage, and estimated costs in timestamped `evals/` directories. Generate JSON reports and Matplotlib comparisons for quality, dimensions, outcomes, latency, cost, and cost versus quality. Include legacy title calls in system costs and track judge costs separately; missing usage or pricing stays unknown. Support separate judging, preserved judgment batches, and offline reporting, reweighting, and compatible-run comparisons.

Legacy evaluation temporarily selects Luna for title generation through the existing site setting, resolving the unsupported Haiku endpoint seen in the initial smoke run. Fixture and model settings are restored afterward, with saved recovery state for interrupted runs. Evaluation conversations remain in the development database.

Real-case scores measure agreement with draft repository references. The new agent's current corpus-retrieval limitation remains visible in the benchmark; the harness does not add corpus text to its prompt.

Refs #179.

## Test plan

- [x] Live Luna smoke run: raw, old, and new each completed the North Dakota fee case and Sol grading.
- [x] Live legacy Luna run: both fictional fee variants completed with Sol grading and title-call usage captured; temporary settings were restored.
- [x] Checked cleanup on timeout, interruption, and controller disconnect; rebuilt reports offline and inspected chart rendering, including critical-failure markers and unknown costs.
- [x] Eval-file lint/format checks, standalone lockfile validation, and `git diff --check` passed.
- [x] Finish and inspect the full baseline: all **252 answers completed**, but grading is still running and has rejected some judge evidence that could not be matched to the answers.
- [X] `make pre-commit` (lint + full application suite) has not been rerun for this eval change.

This internal harness uses manual evaluation checks and has no automated test suite.

---


nicolearagao (dismissed):
LGTM

Two notes on the ND cases, against this rule in CLAUDE.md:

> **Solve directly; escalate to legal aid sparingly.** [...] Don't tell users to
> "get a lawyer" or "call an attorney" [...] 

Both cases escalate at the right moment: `nd-strategy` is the UPL boundary,
`nd-escalation` is safety. `requires_escalation: true` is correct on both.

**1. "or a lawyer" in the answer key** (`cases.yml:153`, `:157`, `:176`)

These lines go into the judge prompt as `expected_facts` and
`acceptable_deferral`. So "talk to a lawyer" scores the same as naming legal aid,
and the rule asks us not to say that. Can we drop "or a lawyer" and keep only
legal aid?

**2. `nd-escalation` may reward deferring when the corpus can answer**

The user is changing their **last name**. The corpus covers that: `waiver.yml:22`
and `:48` say the publication waiver is only for people *not* changing their last
name, and `standard.yml:30-31` give the newspaper notice and the 30-day wait. So
"Can I skip the notice and file?" has a direct answer.

But `acceptable_deferral` calls "acknowledge uncertainty + legal-help off ramp" a
complete answer. That lets an answer skip the procedure and still score well.
Should the expected facts include publication and the 30 days, with the safety
referral added on top?

Neither is blocking. As I non native English speaker, I may be reading the
wording of the rule too strictly. Happy to be corrected on either point.

mitch-mitchel (approved):
updates look good

PR #878, opened 2026-09-11
[867] Remove the DigitalOcean CD workflow from the qa branch

chaco-fl:


mitch-mitchel:
@chaco-fl we actually don't need the merge to QA deploy, I prefer not to have it. It can just be another branch like any other where I trigger the manual deploy targeting that branch (to be clear, I actually prefer it not being special)

mitch-mitchel:
@chaco-fl since `qa` is just another branch like any other at this point, i'm going to just delete and recreate it from `main` since that already has the correct flows on it

Issue #859, opened 2026-09-04
Infra ask: five more Litigant Portal instances (four courts, one QA)


mitch-mitchel:
### Security
**_Confirmed full separate AWS accounts per instance, as well as siloed services_**

- need to lock down any django admin default passwords
- lock down the docker '/interview' admin password and email 
- lock down the dev 'splash' screen on QA when it cuts over
- please audit for standard security items that don't exist in early app dev mode

### What needs to change?

One prod instance is live and working — `deployment/litigant-web`, namespace `litigant`, EKS cluster `courtlistener`, us-west-2, auto-deploying on merge to `main`. It stays as it is.

We need five more environments. **All of them are the same shape as prod and as each other** — one instance serves one court. There is no multi-court or shared mode to build.

| Instance | Facing | Data | Code |
| --- | --- | --- | --- |
| 4 × court partner | partner | persists through internal testing | changes only when we deploy it |
| 1 × QA | internal | wipeable and reseedable on demand | any branch or tag, on demand |

Hostnames are subdomains of `litigantportal.com`, one per court, mapped from the Jurisdictions list. QA at `qa.litigantportal.com`. QA is the same single-court shape as the rest; re-pointing it at a different court's content later is an application concern on our side, not part of this ask.

### No shared services

Each instance gets its own services, with nothing shared between them, so a court is a complete island. Postgres, Redis, and S3 buckets included. This is a business decision on two grounds:

- **Forkability.** A court that wants to take the repo and run it themselves should be able to. An instance whose services are entangled with four others is not something anyone can fork.
- **Cost and impact tracking.** Separate services give us per-court numbers we can use for estimating. A shared pool does not.

Future optimization may find a shape that keeps courts equally safe while sharing AWS resources. That is not part of this ask.

(Bucket names are already per-instance environment variables on our side, so this needs no application change. Sharing a database is not an option regardless: the `Site` row is a singleton holding one court's config.)

#### One exception: docassemble

docassemble is shared across all instances. It is a document-assembly engine rather than a store of court configuration, it is a heavy service (#566, #568), and six copies would cost real money for no isolation gain. A court that forks the repo stands up their own; sharing ours across the instances we host doesn't change that.

One thing that rides with this decision: docassemble does hold interview data, so a shared instance means four courts' answers in one place. That makes #805 (unencrypted `multi_user` storage) and #556 (lock it down for production hosting) more load-bearing, not less. This is not part of this ask.

### Questions for the infra team

[...]

Issue #888, opened 2026-09-11
Decide how docassemble is hosted for QA and the court instances


mitch-mitchel:
## What

Decide the deployment shape for docassemble on FLP infra: which of its supported deployments we run, and whether courts share one instance or each gets its own.

Out of scope: the dev sandbox on the DigitalOcean droplet (#880), which is deliberately temporary and stays all-in-one.

## The options

docassemble supports a single all-in-one Docker container, a docker-compose split (`jhpyle/docassemble-compose`), a Helm chart for Kubernetes, manual non-Docker installation, and a multi-machine cluster. Vendor sizing for a single server is "4 GB of memory and an 80 GB SSD."

## Why the all-in-one does not survive the move

It is the right shape for a dev bench and the wrong one for an environment courts depend on:

- **No rolling update.** One container holds Apache, Postgres, RabbitMQ and celery, so any config change is a full stop and restart.
- **Restore on every boot.** With no external state, the container restores from its backup volume before serving. Observed on the dev box: `initialize` ran for minutes after a `--force-recreate` while the site returned 404.
- **A clean stop is mandatory.** The backup completes only if the container gets its full `stop_grace_period` (600s in our compose). A pod evicted or OOM-killed before that finishes boots from stale state. That is the #701 wipe mechanism, and on a court instance it is data loss rather than an inconvenience.

The split and the Helm chart remove all three, because state lives in real services rather than a self-restoring volume.

## The open decision: shared or per-court

- **Shared.** One docassemble for all courts. Cheaper: one ~20 GB image, one 4 GB floor, one thing to operate. But four courts' interview answers sit in one store, which is the exception to #859's separate-accounts-per-court answer, and it makes #805 load-bearing because `multi_user = True` (required for the prefill handoff) disables per-browser encryption.
- **Per-court.** One docassemble per instance, matching #858 and #859's segregation. Costs a ~20 GB image, a 4 GB floor and a slow boot per court, times four courts plus QA.

## Configuration that follows either way

- `DOCASSEMBLE_BASE_URL` is an internal cluster address. LP reaches docassemble over the cluster network, not the internet.
- `DOCASSEMBLE_PUBLIC_URL` is per-instance, set to the court's own origin, because docassemble builds resume URLs from the host it was called on.
- Each court's ingress mounts docassemble under a path (`/interview/`), never a subdomain, per the README's one-CNAME principle.
[...]

Issue #461, opened 2026-05-28
Deploy production and QA site on CL


nadahlberg:
Infra requirements. We want two deployments, one for the production site and another for QA. For each of these we'll want:
- Pod for django app
- Redis
- Postgres
- Public S3 bucket
- Private S3 bucket

For S3, I think we can just use one pair of buckets for both prod and qa, namespacing the object keys accordingly, instead of requesting 4 buckets.

Other to do items include:
- Get Diego recommendation on postgres specs
- Mitch preps DNS for transfer
- Mitch transfers litigantportal.com to FLP
- Add secret keys to repo for AWS and Docker Hub
- Add CI workflow to autodeploy (see the [MCP workflow ](https://github.com/freelawproject/courtlistener-api-client/blob/main/.github/workflows/deploy.yml)for a minimal example, probably just need to adapt this to handle the QA branch in addition to main).
- system health monitoring - I'm sure you have standard tools for this, so likely obvious
- train Mitch on how to report and file issues - who handles what, critical info for issue/debugging, any easy resets (for example QA pre demo to a court partner where y'all will be busy. Is there a ctrl-alt-del option I can fire)

## Topics / questions for the CL infra meeting

*CI / testing*
- Test stages — adopt CL's fast (DB-free) vs Postgres split for PR CI vs merge gate
- Quality gates — does CL run coverage / mutation testing (`mutmut`/`cosmic-ray`)? Reuse on the engine (`topic_flow/`)?
- Deploy pipeline — adopt CL's workflow (MCP `deploy.yml` ref above) for `main`→prod, `qa`→QA

*Platform (carried from #292 / #294 — now CL's call)*
- Compute / topology — confirm CL's pods/k8s model (per the infra requirements above)
- Container registry — GHCR vs ECR / CL standard
- Secrets — CL's secret manager feeding the `_FILE` convention (closes the #462 gap)
- Postgres (pgvector) specs — Diego's recommendation (also a to-do above)

_Consolidates #292 (CI/CD parent) and #294 (prod deployment), closed in favor of this issue as the single infra/deploy tracker._

nadahlberg:
@mitch-mitchel a few thoughts re the last two you added:
- We can use sentry for error logging. We integrate it into Django, and infra team provisions us a SENTRY_DSN.
- For general health check, I think the infra team adds all the pods to whatever dings the keep-it-up channel
- You can request access from Diego for graphana if you don't have it already. This lets us see the logs.
[...]

PR #900, opened 2026-09-15
Architecture diagrams: today's topology and the per-court target

### LINKED ISSUE #896: System architecture overview diagram (Mermaid / C4)
## What

A maintainable, version-controlled architecture diagram at the C4 Container level — the players and how they plug together end to end — as Mermaid source in `docs/architecture/system-overview.md`. Renders in GitHub / Obsidian / VS Code and exports to PNG/SVG for decks. Color-codes ours vs. court content vs. external, and makes the three shared surfaces (corpus/KB, case-fact/Briefcase state, forms) unmissable as the single source of truth.

## Why

No single view shows how the parts — Django app, AI-free topic flow, AI assistant + tooling, docassemble, corpus/KB, forms, EKS, court partners, users — connect and where they share. New devs need it to onboard; stakeholders need the end-to-end story. Mermaid keeps it dev-editable and diffable, not a bespoke artifact only one tool can update.

## Definition of done

- [ ] `docs/architecture/system-overview.md`: Container-level Mermaid diagram + legend + per-node sub-lists
- [ ] Three shared surfaces visually distinct; ours / court / external color-coded; renders on GitHub
- [ ] Component-level zoom-ins (per node) scoped as follow-on issues, not required here

mitch-mitchel:
Splits the architecture picture in two, because one diagram was claiming the other's state: `system-architecture.drawio` is now what actually runs, and `target-isolation.drawio` is the #859 per-court-account target, banner and all.

The correction that drove it: the boundary read "Per-court AWS account", but `deploy.yml` pins `EKS_CLUSTER_NAME: courtlistener` and `EKS_NAMESPACE: litigant`. The portal is one namespace in FLP's shared cluster, with QA as a second namespace beside it. A court's tech resource reading the old boundary would have concluded their data was already isolated.

Other things the diagram asserted that the repo contradicts:

- images publish to Docker Hub, not ECR
- docassemble is the stock `jhpyle/docassemble` image with its own database, so it gets no arrow to the app's Postgres
- there is no `SESSION_ENGINE`, so sessions are in Postgres and Redis is the cache only — the opposite of the obvious guess
- LiteLLM is a library inside the app, not a service in the path
- third-party court/state services are links the litigant follows off-site, not integrations
- in the EKS breakout, the cache edge collapsed into the Postgres icon and read as Postgres → Redis

Added the two nodes a court tech asks about first and the diagram could not answer: S3 (private bucket for uploads, public for static/media) and Secrets Manager. Every node now carries what it does, not just what it is. The diagram also carries its own legend, and court-authored content has its own colour.

Note on history: f1dfda5 introduced the "Per-court AWS account" label and 82f0fbf takes it back out. That was the right call on what we knew; the deploy workflows said otherwise.

Built in draw.io rather than the Mermaid named in the issue — rationale in `docs/architecture/README.md`: mermaid gives clean layout or AWS icons, never both. The shared-surfaces view the issue also calls for is a content-altitude diagram rather than a deployment one, so it ships as its own issue: #901.

Closes #896

Test plan:

- [ ] The three `.drawio.svg` files render as images on GitHub
- [ ] Each `.drawio.svg` reopens editable in draw.io
- [ ] `drawio -x -f svg -e -o <n>.drawio.svg <n>.drawio` reproduces the committed renders
- [ ] No label overlaps a box border or another label at 100%


docs/architecture/system-architecture.drawio
```diff
@@ -0,0 +1,193 @@
...
+        </mxCell>
+        <mxCell id="ingress" value="Ingress&#10;TLS termination + site-wide access gate" style="rounded=1;whiteSpace=wrap;html=1;fillColor=#FFFFFF;strokeColor=#232F3E;fontColor=#232F3E;fontSize=11;" vertex="1" parent="1">
+          <mxGeometry x="70" y="400" width="150" height="90" as="geometry" />
+        </mxCell>
+
+        <mxCell id="app" value="Django app — our code&#10;(Litigant Portal + Topic Flows + AI assistant + agent tooling)&#10;deployment: litigant-web" style="rounded=1;whiteSpace=wrap;html=1;fillColor=#E6EEFB;strokeColor=#2F5FB0;fontColor=#12325F;fontSize=11;verticalAlign=middle;align=center;" vertex="1" parent="1">
+          <mxGeometry x="370" y="240" width="320" height="95" as="geometry" />
+        </mxCell>
+
+        <mxCell id="da" value="docassemble — guided interviews + document assembly&#10;the stock jhpyle/docassemble image, pulled and configured&#10;not our code · own container, own database" style="rounded=1;whiteSpace=wrap;html=1;fillColor=#F2F2F2;strokeColor=#666666;fontColor=#333333;fontSize=11;verticalAlign=middle;align=center;dashed=1;" vertex="1" parent="1">
```
nicolearagao, line 51: docassemble is drawn inside the prod litigant namespace. I do not think it runs there yet.

docs/docassemble.md:40 says its production home comes with the CL infra move (
#461), and that QA hosting is only a temporary demo host. In this repo docassemble appears in two places: local compose, and deploy/qa-do/docker-compose.qa.yml:109. Issue 
#888 is still open about where it will be hosted.

This is the same problem as the old "Per-court AWS account" label. A court reader would think the interview part is already live on AWS.

Can we draw it with a dashed border and add a note like "not deployed yet, see #888"? Or move it outside the namespace box.

docs/architecture/system-architecture.drawio
```diff
@@ -0,0 +1,193 @@
...
+        <mxCell id="g_delivery" value="Delivery" style="rounded=1;whiteSpace=wrap;html=1;fillColor=#F0F7FF;strokeColor=#147EBA;dashed=1;verticalAlign=top;align=left;spacingLeft=12;spacingTop=8;fontColor=#147EBA;fontSize=13;fontStyle=1;" vertex="1" parent="1">
+          <mxGeometry x="300" y="770" width="1150" height="190" as="geometry" />
+        </mxCell>
+
+        <!-- ===== Nodes ===== -->
+
+        <mxCell id="browser" value="Litigant browser" style="rounded=1;whiteSpace=wrap;html=1;fillColor=#FFFFFF;strokeColor=#232F3E;fontColor=#232F3E;fontSize=12;" vertex="1" parent="1">
+          <mxGeometry x="85" y="250" width="120" height="60" as="geometry" />
+        </mxCell>
+        <mxCell id="ingress" value="Ingress&#10;TLS termination + site-wide access gate" style="rounded=1;whiteSpace=wrap;html=1;fillColor=#FFFFFF;strokeColor=#232F3E;fontColor=#232F3E;fontSize=11;" vertex="1" parent="1">
```
nicolearagao, line 43: The ingress box says "TLS termination + site-wide access gate". TLS is correct. But the access gate is SitePasswordMiddleware inside Django (settings.py:73, app/middleware.py:9).

Three things follow from that:

it only runs when SITE_PASSWORD is set
it skips /api/, static and media
/interview/* goes to docassemble without passing Django, so interviews are not gated at all
So it is not at the edge, and it is not site-wide. Putting it on the ingress tells a court reader that the edge protects everything.

Can we remove it from the ingress label, and add it to the Django app box instead? For example: "temporary site password gate (pages only, not /api/)".

docs/architecture/README.md
```diff
@@ -0,0 +1,70 @@
+# Architecture diagrams
```
nicolearagao, line 1: Nothing links to this folder. docs/README.md is the docs index, and it has no Architecture section. People will not find these diagrams.

Can this PR add a section there?

docs/architecture/eks-breakout.drawio
```diff
@@ -0,0 +1,132 @@
...
+          <mxGeometry x="470" y="306" width="260" height="88" as="geometry" />
+        </mxCell>
+        <!-- small K8s Deployment badge on the web box -->
+        <mxCell id="web_badge" value="" style="sketch=0;html=1;dashed=0;whiteSpace=wrap;aspect=fixed;fillColor=#2875E2;strokeColor=#ffffff;shape=mxgraph.kubernetes.icon2;prIcon=deploy;" vertex="1" parent="1">
+          <mxGeometry x="478" y="313" width="26" height="26" as="geometry" />
+        </mxCell>
+
+        <!-- Migration Job: ephemeral temp pod (K8s Job icon) -->
+        <mxCell id="mig" value="Migration Job (temp pod)&#10;collectstatic + migrate, deleted after" style="sketch=0;html=1;dashed=0;whiteSpace=wrap;aspect=fixed;verticalLabelPosition=bottom;verticalAlign=top;align=center;fillColor=#2875E2;strokeColor=#ffffff;fontColor=#232F3E;fontSize=11;shape=mxgraph.kubernetes.icon2;prIcon=job;labelWidth=140;" vertex="1" parent="1">
```
nicolearagao, line 47: This is called "Migration Job", but it is not a Kubernetes Job. It is a plain Pod (kubectl run ... --restart Never, deploy.yml:136).

The difference matters here. deploy.yml:125-127 explains that a plain Pod has no TTL, and that is why the deploy has a reap step. A Job would clean itself up.

Can we call it "Migration Pod (temp, reaped on next deploy)"?

docs/architecture/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+- `system-architecture.drawio` — **today.** What is actually deployed right now.
+- `target-isolation.drawio` — **target (#859).** One AWS account per court. Not built yet.
+- `eks-breakout.drawio` — component-level zoom into a court's instance internals.
+
+Each has a committed `.drawio.svg` render alongside it. That SVG shows as an image on GitHub and in Obsidian **and** reopens editable in draw.io (the diagram XML is embedded) — it's what people view without opening the app.
+
+## Editing
+
+1. Install draw.io: `brew install --cask drawio`.
```
nicolearagao, line 17: brew install --cask drawio only works on macOS. Some of us are on Linux.
Instead of listing distros, can we link the releases page? It has .deb, .rpm and AppImage in one place:

```
1. Install draw.io:
   - macOS: `brew install --cask drawio`
   - Linux: download a package from https://github.com/jgraph/drawio-desktop/releases
```

nicolearagao (changes requested):

PR #903, opened 2026-09-15
Run dependabot monthly, and cover uv and pre-commit

### LINKED ISSUE #902: Align dependabot with the org: monthly, plus uv and pre-commit
### What needs to change?

- Move the existing `github-actions` block from weekly to monthly
- Add `package-ecosystem: uv` for `pyproject.toml` + `uv.lock`
- Add `package-ecosystem: pre_commit` for `.pre-commit-config.yaml` hook revisions
- Group each into one PR; ignore semver-patch on `uv`

### Why?

Python dependencies and pre-commit hook revisions get no version updates today — the config covers `github-actions` only, so both drift until someone bumps them by hand. Advisory-driven security updates still fire, so this is about staleness, not exposure.

Monthly is the org convention: courtlistener, juriscraper, x-ray, courts-db, reporters-db and recap-chrome all run monthly. LP's weekly is the outlier, and the faster feed isn't being consumed anyway — #891 is a one-line bump that has sat unreviewed since 9/14. Less noise, same coverage.

### Definition of done

- [ ] `.github/dependabot.yml` covers `github-actions`, `uv` and `pre_commit`, all monthly
- [ ] `uv` ignores `version-update:semver-patch`; both new ecosystems grouped into one PR each

mitch-mitchel:
The config covered `github-actions` only, so Python dependencies and pre-commit hook revisions got no version updates at all — both drifted until someone bumped them by hand. Advisory-driven security updates fire regardless, so this was staleness, not exposure.

Adds `uv` (`pyproject.toml` + `uv.lock`) and `pre-commit` (`.pre-commit-config.yaml` hook revs), each grouped into a single PR. `uv` ignores semver-patch bumps across the tree — more noise than signal, and an advisory still comes through the filter.

Moves `github-actions` from weekly to monthly to match. Monthly is the org convention — courtlistener, juriscraper, x-ray, courts-db, reporters-db and recap-chrome all run monthly. LP's weekly was the outlier, and the faster feed was not being consumed anyway.

Per GitHub's supported-package-managers table, `pre-commit` is version-updates only: no security-update support, and no private-repository support. Both are fine here — this repo is public, and security updates for Python come through `uv`.

Closes #902

Test plan:

- [ ] `.github/dependabot.yml` parses as valid YAML
- [ ] GitHub reports no Dependabot config errors on the repo after merge
- [ ] Insights → Dependency graph → Dependabot lists all three ecosystems


nicolearagao (dismissed):
LGTM

PR #911, opened 2026-09-16
[881] Adding Wipe DataBase Step to QA

### LINKED ISSUE #881: QA deploys rebuild the database from the deployed branch
## What

Every EKS deploy to QA starts from an empty database and builds it from the branch being deployed: drop, `migrate`, seed, serve. No migration state carried between deploys, no conflict resolution, no hard fails to unblock. Deploying an older branch is the same operation as deploying a newer one.

In the deploy pipeline, not only at pod startup:

- drop and recreate the database
- `manage migrate --noinput`
- `manage sync_corpus` with `CORPUS_COURT` set for QA
- `manage bootstrap_superuser`

## Seed source of truth: the deployed branch

The seed is whatever that branch carries: `litigant_portal/corpus/` via `sync_corpus`, the `post_migrate` receivers (`ensure_site_row`, `ensure_permission_groups`), and `bootstrap_superuser` from env. Nothing persists by being entered into QA's admin. Anything QA needs after a wipe goes into the seed.

## Why

QA's job is testing arbitrary branches. Migration state surviving a deploy is what makes that unsafe, so branches get avoided or QA gets hand-repaired. `seed_data` was removed by `936b7cc`, so the corpus tree is already the only supported seed path.

## Gaps a full wipe exposes

- `Site.fast_model` and `Site.assistant_model` are not in `court.yml` and are not seeded, so every deploy resets the chat model to `DEFAULT_BEDROCK_MODEL`. Testing a non-default model dies silently.
- Test and persona accounts vanish each deploy, which makes #464 load-bearing for QA.
- Chat threads and transcripts vanish, so #802's end-to-end audit has to fit inside one deploy window.
- Uploaded media lives outside the database, so a wipe orphans the files.

## A malformed corpus already fails the deploy, and should

No flag is needed for this. A corpus that does not load raises during load, `sync_corpus` turns that into a `CommandError`, and the step exits non-zero. Ahead of that, `apps.py` registers `checks.corpus` and `topic_flow.checks` as Django system checks and `migrate` carries `requires_system_checks = "__all__"`, so a bad corpus fails at migrate before sync even runs. Two automatic gates.

That is the behavior we want. A validation failure means the branch's corpus is malformed or missing fields, not that someone disagrees with its content. And no QA scenario is corpus-free: the home grid, every flow page and the chat's topic grounding all read `Topic` rows, so a branch whose corpus will not load has nothing to QA.

`--strict` is a different lever than its name suggests: "Also delete rows the corpus no longer names." On a database dropped and rebuilt every deploy there are no stale rows, so it is close to a no-op here. Kept because it is the correct setting if a wipe is ever skipped.

## Definition of done

- [ ] QA deploy of any branch, newer or older, succeeds with no manual database work
- [ ] Database dropped and rebuilt as a deploy step, before migrate
- [ ] `sync_corpus` and `bootstrap_superuser` run in that pipeline, not only at pod startup
- [ ] `CORPUS_COURT` has an explicit QA value
- [ ] Site model selections seeded, or reset-to-default accepted explicitly
- [ ] Written down: QA data persists only through the branch's seed

chaco-fl:
close #881


.github/workflows/qa-deploy.yml
```diff
@@ -7,6 +7,11 @@ on:
         description: 'Branch or tag to build and deploy to QA'
         required: true
         default: 'main'
+      reset_db:
+        description: 'Drop every table and rebuild the database from the deployed branch'
+        type: boolean
+        required: false
```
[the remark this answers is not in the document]
mitch-mitchel: this is by design for what we are doing right @chaco-fl

.github/workflows/qa-deploy.yml
```diff
@@ -95,6 +100,22 @@ jobs:
...
+
+          cursor = connection.cursor()
+          cursor.execute("SET lock_timeout = 30000")
+          cursor.execute("SELECT tablename FROM pg_tables WHERE schemaname = current_schema()")
+          tables = [row[0] for row in cursor.fetchall()]
+          with transaction.atomic():
+              for table in tables:
+                  cursor.execute("DROP TABLE IF EXISTS " + connection.ops.quote_name(table) + " CASCADE")
+          print("Dropped", len(tables), "tables")
+          PY
```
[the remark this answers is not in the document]
mitch-mitchel: The gap the bot is pointing at is the window after it succeeds methinks

 QA has no uptime requirement, so this is cheap to close — scale litigant-web to 0 before the drop and back up after Bootstrap Superuser. Worth doing, since the alternative is a QA that looks broken to anyone who happens to be on it mid-deploy.

.github/workflows/qa-deploy.yml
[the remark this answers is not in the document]
chaco-fl: yes

.github/workflows/qa-deploy.yml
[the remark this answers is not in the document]
chaco-fl: I have added the the step do scale to 0

.github/workflows/qa-deploy.yml
```diff
@@ -95,6 +100,22 @@ jobs:
           echo "collectstatic failed--aborting deploy"
           exit 1
 
+      - name: Reset Database
```
mitch-mitchel, line 112: A failure here reports as a migration failure. `Handle Migration Error` (line 121) carries a bare `if: failure()`, so when this step fails, `Apply Migrations` is skipped and that handler still runs — printing "Migrations failed. Open a shell into pod temp-pod-… to investigate." Whoever is debugging a lock timeout on the drop gets pointed at migrations.

Two ways out: give this step its own `Handle Reset Error`, or give `Apply Migrations` an `id:` and scope the existing handler to `if: failure() && steps.<id>.conclusion == 'failure'`. The second is less to maintain and fixes the same latent issue on `Handle Collectstatic Error`.

.github/workflows/qa-deploy.yml
```diff
@@ -88,22 +93,58 @@ jobs:
...
           exit 1
 
+      - name: Sync Corpus
+        run: kubectl exec -n ${{ env.EKS_NAMESPACE }} temp-pod-${{ needs.build.outputs.sha_short }} -- manage sync_corpus --strict
+
+      - name: Bootstrap Superuser
+        run: kubectl exec -n ${{ env.EKS_NAMESPACE }} temp-pod-${{ needs.build.outputs.sha_short }} -- manage bootstrap_superuser
+
+      - name: Scale Up litigant-web
+        if: ${{ always() && steps.scale_down.outcome == 'success' }}
```
mitch-mitchel: Gating on `outcome == 'success'` means a partial scale-down leaves QA at zero replicas. If `--replicas=0` succeeds and the `kubectl delete pod` below it times out, this step's outcome is `failure`, so Scale Up never runs and the site stays down until someone notices. `always() && steps.scale_down.conclusion != 'skipped'` covers the case the `always()` was there for.

.github/workflows/qa-deploy.yml
```diff
@@ -88,22 +93,58 @@ jobs:
...
           echo "collectstatic failed--aborting deploy"
           exit 1
 
+      - name: Scale Down litigant-web
+        id: scale_down
+        if: ${{ inputs.reset_db }}
+        run: |
+          echo "WEB_REPLICAS=$(kubectl get deployment/litigant-web -n ${{ env.EKS_NAMESPACE }} -o jsonpath='{.spec.replicas}')" >> $GITHUB_ENV
+          kubectl scale deployment/litigant-web -n ${{ env.EKS_NAMESPACE }} --replicas=0
+          kubectl delete pod -n ${{ env.EKS_NAMESPACE }} -l service=litigant-web --ignore-not-found --timeout=120s
```
mitch-mitchel, line 110: This is what makes the drop wait for the old pods to actually go, so the selector has to be right. If the deployment's pods don't carry `service=litigant-web`, it matches nothing, returns immediately, and the drop races pods that are still terminating and still holding connections — the original problem, hidden behind a step that passes. No manifests in this repo, so only you can confirm the label.

`kubectl wait --for=delete pod -l <selector> --timeout=120s` is the more explicit form and fails loudly on a bad selector.

mitch-mitchel:
One housekeeping note: the PR body is empty, so there is no closing keyword — #881 will stay open after this merges. Adding `Closes #881` to the description takes care of it, and keeps the board automation moving the issue to Done on its own. 

edit (mitch) sorry claude is ai-splaning here

.github/workflows/qa-deploy.yml
chaco-fl: Fixed in 9f8c851

.github/workflows/qa-deploy.yml
chaco-fl: Confirmed against the live deployment: `spec.selector.matchLabels` is    `{"service": "litigant-web"}`

mitch-mitchel (approved):
LGTM - thanks !

Issue #908, opened 2026-09-15
Wire QA docassemble config for the settings-only handoff


nicolearagao:
### What needs to change?

Once #912 lands, the portal builds every docassemble URL from settings and the
corpus carries only a package reference (`docassemble.ndnamechange:...`).
docassemble is now live on EKS QA (#550) and the `ndnamechange` package is
installed on it (2026-09-18), so **QA is what this issue wires**. Three things:

- **Infra:** add three keys to the `litigant-env` secret in AWS Secrets Manager
  (`k8s/qa-litigant/...`), so the web pods pick them up on the next deploy or
  restart (the deploy workflow already re-syncs the secret every run):
  - `DOCASSEMBLE_BASE_URL` — the internal service address, including the
    `/interview` path prefix. Needs the Service name and port confirmed from
    `qa-litigant/5_docassemble.yml`; expected shape: `http://docassemble/interview`.
  - `DOCASSEMBLE_PUBLIC_URL` — `https://qa.litigantportal.com/interview`
  - `DOCASSEMBLE_API_KEY` — the `DA_ADMIN_API_KEY` value from Bitwarden
    (bridge value; a dedicated lower-privilege key is a follow-up, see below)
- **Deploy QA** after #912 merges, so the handoff runs against the installed
  package instead of the old Playground URL.
- **Verify end to end:** the ND name change flows show the interview button,
  the POST handoff lands inside the interview with the reviewed answers
  prefilled, and startup is clean (no registry "no docassemble configured"
  warning, no `docassemble.W001`).

### Why?

QA's web pods get no `DOCASSEMBLE_*` env vars today, so the handoff opens the
interview **unprefilled** (the code's deliberate fallback when no API key is
configured). And once #912 merges, an environment with no `DOCASSEMBLE_*` at
all hides the interview button entirely — so without this wiring, QA loses the
handoff instead of completing it. This is the last gap in the LP → docassemble
chain on QA and blocks the #585/#587 verification of the full flow.

### Definition of done (optional)

- The ND name change flows on QA show the interview button and the prefilled
  handoff works end to end against the installed `docassemble.ndnamechange`
  package (no Playground reference involved).
- Startup is clean on the QA deploy: no registry warning, no `docassemble.W001`.
- Out of scope, tracked elsewhere: dev-box wiring (#880), a dedicated
  low-privilege API key and prod wiring (#556), automated package installs
  (separate issue to file).

mitch-mitchel:
[...]

Issue #587, opened 2026-07-01
QA on AWS: load app secrets and verify full working state on latest code


mitch-mitchel:
### What

`qa.litigantportal.com` is now serving from the new AWS environment infra stood up 2026-06-30. Two parts:

1. **Load the LP app config/secrets into the AWS env** — the prod-profile `.env` set: `SECRET_KEY`, `DOMAIN`, `ALLOWED_HOSTS`, `DEPLOYMENT_ENV=qa`, `CHAT_ENABLED`, `CHAT_MODEL`, `OPENAI_API_KEY`, `POSTGRES_PASSWORD`.
2. **Full verification pass — not just secrets.** Confirm the environment is running latest `main` and everything works end-to-end, not merely booted.

### Why

QA is the shared surface Jessica and other FLP teams see, and it's the target Jessica's QA scripts (#585) run against. It needs to be on latest code and fully functional before that QA round — a booted-but-stale or half-configured env would invalidate the testing.

### Definition of Done

- [ ] App secrets/config present in the AWS env
- [ ] Environment running latest `main` (image tag / commit matches HEAD)
- [ ] HTTPS serving, cert valid (Caddy or AWS-fronted — confirm which)
- [ ] `migrate` + `collectstatic` ran; static assets serve
- [ ] `/interview/` docassemble route live (guards the #558 regression)
- [ ] AI chat streams end-to-end (confirms `OPENAI_API_KEY` wired)
- [ ] ND name-change Topic Flow works on both tracks: fact-gather → packet → official form-PDF links + resource links render
- [ ] Deploy path confirmed: does merge-to-`main` CD reach the AWS env? Note what changed vs. the DigitalOcean path (`QA_HOST`/`QA_SSH_KEY`, GHCR vs Docker Hub image source)
- [ ] Update `docs/QA-DEPLOY.md` + CLAUDE.md to reflect AWS (currently document a DigitalOcean VPS)

mitch-mitchel:
@ERosendo @chaco-fl @nadahlberg had a good fight with Claude, and confirmed the clean github flow, with optional actions to deploy to other environments. I like this best - `staging.litigantportal.com` or `qa.litigantportal.com` can exist as environments but we trigger deploys to them without requiring the bookkeeping of branch merges and restricted flows.  PS - this mirrors courtlistener and only adds the `workflow_dispatch` bit.


## Branching & deploy strategy (from infra chat, 2026-07-01)

[...]

PR #912, opened 2026-09-16
feat(topic_flow): docassemble launch URL and API root become validated configuration

### LINKED ISSUE #879: docassemble launch URL and API root become validated configuration
## What

The docassemble API root falls back to the host named in the corpus `interview_url` whenever `DOCASSEMBLE_BASE_URL` is unset (`services/docassemble.py`, `_target()`), and the shipped corpora name a QA host plus a Playground reference (`docassemble.playground1:petition-standard.yml`).

Make the API root come only from configuration, and have the corpus carry the interview *reference* rather than a full URL plus host. The loader validates the reference; the environment supplies where to send it.

## Why

A value in author-controlled content currently decides where the API key and the litigant's answers are sent. Courts author this content later, so that boundary needs validating rather than trusting — the repo's open-contracts stance applied to the one surface that carries a credential.

A Playground reference also isn't deployable outside the authoring user's project, so the corpus that ships to prod names something only QA can resolve.

PR #869 takes the fail-closed guard (refuse to run when a key is set without a base URL). This is the design change behind it.

## Definition of done

- [ ] API root comes from configuration only; no host is derived from corpus content
- [ ] Corpus carries an interview reference, validated by the loader
- [ ] Launch URL is resolved per environment, so prod never names a Playground project
- [ ] Both ND name-change corpora migrated off the hardcoded QA URL
- [ ] Drift guard still pins the mapped variables against the versioned interviews

nicolearagao:
## What changed

Closes #879.

Every docassemble URL is now built from settings; the corpus carries only a
validated interview reference (`docassemble.ndnamechange:data/questions/...`).
Author-controlled content can no longer decide where the API key and the
litigant's answers are sent — the open-contracts boundary #879 asked for,
completing the fail-closed guard from #869.

## Deploy consequence — decided in the open

Merging this hides the "Fill out your forms" button on any environment
without `DOCASSEMBLE_*` configured. Concretely:

- **QA (EKS)** has no docassemble (#550), so the button disappears there —
  but today's button already points at `qa.litigantportal.com/interview/`,
  which 404s on EKS. This trades a broken button for a hidden one plus a
  startup log naming the affected flows. Known gap until #550 + the QA half
  of #908 land. Ordering is tracked in #587.
- **The dev box** runs docassemble but its compose passes no
  `DOCASSEMBLE_*` vars into the web service; #908 (retargeted at dev) wires
  them and installs the interviews as a real package.

## Test plan

- [x] Manual, against the local bench (`make docassemble-up`, interviews
  installed as the `ndnamechange` package, `DOCASSEMBLE_*` set): the packet
  button renders, prefill lands inside the interview; unset both URLs,
  restart, button gone and registry warning names both ND flows; key set
  with base URL unset, `manage.py check` reports `docassemble.W001`.


mitch-mitchel:
Reviewed at `879-docassmble-validated-config`. The shape is right: content no longer decides the host, and the fail-closed paths (404 when unconfigured, redirect to plain on error, hidden button) stay consistent across the renderer, the view, and the CSRF fallback. Nothing here blocks.

One substantive item, worth a look before #550 and #908 land:

**`services/docassemble.py:94` — `_public()` drops the path prefix.** It keeps only scheme and netloc, while `interview_launch_url()` (line 42) uses the full base including its path. With the QA config we're standing up:

```
DOCASSEMBLE_BASE_URL=http://docassemble
DOCASSEMBLE_PUBLIC_URL=https://qa.litigantportal.com/interview
```

the plain launch link keeps `/interview`, but a prefilled resume URL loses it. `test_resume_url_is_rewritten_onto_the_public_origin` asserts that drop as expected behavior. Whether it actually 404s depends on docassemble's own url root config, which I haven't verified. If it does, the failure shape is unkind: the unprefilled fallback still works, so only litigants who completed a flow hit the dead link. The manual test plan covers the local bench, where there's no prefix to lose.

**PR body vs. behavior:** the description mentions "a startup log naming the affected flows," but `registry.load()` is lazy and nothing calls it from `apps.ready()` or `manage.py check`. The warning first fires on whichever request touches the registry, once per worker, so it won't appear in deploy logs where an operator would look for it.

Nits, take or leave:

- `topic_flow/schema.py:42` — the path pattern permits `..`, so `docassemble.pkg:data/questions/../../../secret.yml` validates. It can't redirect the key now the host is settings-only, but rejecting `..` is cheap given this module is the validated-contract boundary.
- `topic_flow/registry.py:68` — `_warn_unservable_handoffs` re-implements the configured-ness rule instead of calling `interview_launch_url`. If the launch rule changes, the warning drifts from what the page actually renders.
- `docs/docassemble.md:80` — install step 3 says the server resolves `docassemble.ndnamechange:petition-standard.yml`, which is the short alias that line 58 and the schema both reject.

mitch-mitchel (approved):
LGTM

